PR #4, opened 2018-05-09
Add initial batch line lists

tepickering:
This PR adds an initial batch of line lists to use for wavelength calibration. One batch is taken directly from IRAF's ``noao`` package, the other is taken from NIST's ASD (https://www.nist.gov/pml/atomic-spectra-database-contents). The IRAF lists include both air and vacuum wavelengths. The files have been sorted accordingly into separate directories here.


jehturner:
Great :+1:. Would it be helpful to have those in a standard format? I imagine they'll get loaded into some fixed abstraction in Python, but the issue may be consistency for users who want to edit a copy (other than programatically)? Just a thought.

PR #32, opened 2018-09-27
Adding table parsing utils for NIST master table (dotAstronomy hack day)

mfixstsci:
These utils can be used to sort the master line list based on elements or wavelengths. Hoping that users will be able to import them into an `ipython` session and use them to parse, build, and save table for their specific science cases.


mfixstsci:
Hey @tepickering, that was an accident! But I think that would be okay!

Issue #27, opened 2018-06-15
Define format for line lists


tepickering:
The line lists added in #4 are taken verbatim from their sources and are thus somewhat heterogeneous. We should define a consistent format that is easily parseable and contains the necessary metadata like wavelength units and whether they're air or vacuum.

PR #33, opened 2018-09-27
Added more NIST lists (.Astro Hack Day)

lprichard:
Added all the air and Ne, Kr, Xe, Si vac lists for strong lines. Lists are in different format to original lists delimited by 2+ spaces rather than commas. Copied tables from [https://physics.nist.gov/PhysRefData/Handbook/element_name_a.htm](https://physics.nist.gov/PhysRefData/Handbook/element_name_a.htm). Simpler format than the output of astroquery.nist.query() with all the information needed


lprichard:
Done at .Astro Hack Day

mfixstsci:
I was hoping @tepickering could help shed some light on this! @lprichard noticed that the strength column for some of these strong line lists have notations that include commas (i.e. P,r). So as a result, there are no commas separating the values in the lists for this PR because some of the rows would have an extra column because of this notation.

Did you see or deal with this at all for the lists you committed @tepickering ? I am only asking because I am concerned #31 or #32 won't play nicely with comma and white space delimited files at the same time.

tepickering:
formatting of the line list data is an open issue, #27. core question is do you massage the data files or wrap them with methods to do the parsing? we discussed that a bit at pyastro18, but never settled on a path. 

that all said, though, i think any data files should at least be able to be parsed by ``astropy.table`` or ``astropy.ascii`` with minimal wrangling.

mfixstsci:
hey @tepickering thanks for the information! I will go ahead and look into wrangling these txt files.

lprichard:
I grabbed the data straight from NIST and there the info is separated by spaces (more than one so as not to separate atoms from their states e.g. Ar I). I reckon going forward the best format for new files is columns separated by a tab or few spaces. The ones that were originally in the directory will therefore need to be changed, this is only 5 files though. Then Mees' code can wrangle that standard format into a master txt file with added columns (bool and vac/air etc.) that we discussed would be needed and this output catalogue can be read by astropy.table/ascii. As long as we document the required format (intensity, wavelength, species, ref from NIST) of any additional lists users may want then these can easily be added to the master, this seems like the simplest solution to keep it as general and adaptable as possible. Alternatively, we continue to get a complete directory of lists and put them all in a master from the get go, this may get to be a big task though if we start including molecules..

PR #38, opened 2018-12-05
First test

crawfordsm:
This adds a test to the repository for build_nist_table but also sets up the infrastructure for using the datasets repository in the testing


tepickering:
the unwatched travis finally ran. looks good. merge away!

PR #41, opened 2019-08-02
created empty identify_target function

simontorres:
The goal of this pull request is to implement a function to identify the location of the spectrum of a point source in a 2D image.

### DIFF
--- a/specreduce/target_identify.py
+++ b/specreduce/target_identify.py
@@ -1,0 +1,30 @@
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+
+import logging
+import numpy as np
+from scipy.stats import sigma_clip
+
+
+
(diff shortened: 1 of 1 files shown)


simontorres:
the initial commit does not contain much code since pyastro19 has arrive to an end and I decided to put the PR already. Here is a document that describes what I'm going to implement. https://github.com/simontorres/general_documentation/blob/master/target_identification/target_identification_test.ipynb

jehturner:
Is this of interest? I realize it may be a bit tied to our own data access class.

https://github.com/GeminiDRSoftware/DRAGONS/blob/spectroscopy/gempy/library/tracing.py

stscicrawford:
@jehturner That does look very helpful and has many of the underlying features that would be needed.

PR #45, opened 2020-03-22
Add draft of flux calibration functionality to specreduce

### LINKED ISSUE #10: flux calibration of the spectra
- [ ] Compare a standard star observations to a flux calibration standard to compute the correction
- [ ] Apply the correction to observed science observations

kakirastern:
Fixes #10. 

This PR addresses the issue of missing code for performing flux calibration on a 1D spectrum. Using `pydis` as a reference for now, but will need to make drastic changes to it to cater to the specific needs of the Astropy community, to adhere more to the `gemini` standard. 

References: 
* `genini` link [here](https://github.com/GeminiDRSoftware/DRAGONS/blob/master/geminidr/core/primitives_spect.py)
* `pydis` link [here](https://github.com/TheAstroFactory/pydis/blob/master/pydis/pydis.py)


setup.cfg
```diff
@@ -10,6 +10,11 @@ long_description = file: README.rst
 long_description_content_type = text/x-rst
 edit_on_github = False
 github_project = astropy/specreduce
+install_requires = astropy, numpy, scipy, matplotlib, specutils
+# version should be PEP440 compatible (http://www.python.org/dev/peps/pep-0440)
+version = 0.0.dev
+minimum_python_version = 3.6
+
```
tepickering, line 17: it looks like this still needs to be changed to match APE17. specifically, `version` shouldn't be here and the stuff other than `astropy` are probably more appropriately put under `options.extra_requires`. i also strongly suggest we set the minimum python version to 3.7 so that we can use data classes where appropriate.

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
 # This module implements the base CCDPROC functions
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
```
tepickering, line 4: general comment for this whole file: i would suggest putting flux calibration specific utilities in their own file, e.g. `fluxcal.py` or `flux_calibration.py`. `mag2flux()` may be appropriate to stay and be made public.

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+        if mode == "linear":
+            sensfunc = np.interp(obj_wave, obj_wave_ds, logsensfunc)
+        elif mode == "spline":
+            spline = UnivariateSpline(obj_wave_ds, logsensfunc, ext=0, k=2, s=0.0025)
+            sensfunc = spline(obj_wave)
+        elif mode == "poly":
+            fit = np.polyfit(obj_wave_ds, logsensfunc, polyorder)
+            sensfunc = np.polyval(fit, obj_wave)
+
+        if show is True:
```
tepickering, line 171: a better pattern for providing graphical output is to bundle the plot into a `matplotlib.figure.Figure` object that is then returned to the caller. how figures are handled will vary depending on context (interactive vs non-interactive, which backend is used, etc) so we don't want to bake assumptions into core code here.

setup.cfg
kakirastern: Thanks, just checked out the doc at https://github.com/astropy/astropy-APEs/blob/master/APE17.rst and have updated accordingly.

pllim:
FWIW, all the flux conversion code already exists, either directly in `astropy.units` or in https://github.com/spacetelescope/synphot_refactor/blob/master/synphot/units.py . I have issue spacetelescope/synphot_refactor#156 to move the unit handling stuff out as its own thing but gotten no interest so far. 🤷‍♀

kakirastern:
> FWIW, all the flux conversion code already exists, either directly in `astropy.units` or in https://github.com/spacetelescope/synphot_refactor/blob/master/synphot/units.py . I have issue [spacetelescope/synphot_refactor#156](https://github.com/spacetelescope/synphot_refactor/issues/156) to move the unit handling stuff out as its own thing but gotten no interest so far. 🤷‍♀

@pllim Let me look into the issue at https://github.com/spacetelescope/synphot_refactor/issues/156 and see if I can do a follow-up for it.

specreduce/core.py
kakirastern: Agreed

specreduce/core.py
jradavenport: FWIW that's how I've been structuring my version of this exact same update

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+        msg1 = "Error: No 'onedstd' standard star file has been specified." \
+               "Cannot proceed further."
+        raise ValueError(msg1)
+
+    std = std.lower()
+    std_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                           "datasets", "onedstds")
+
+    if os.path.isfile(os.path.join(std_dir, std)):
+        std_wave, std_mag, std_wth = np.genfromtxt(
```
jradavenport, line 115: I've been switching to using astropy tables for reading these onedstds in, hoping that's more stable and useful?

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+# from astropy.wcs import WCS
+
+import numpy as np
+
+# from scipy.optimize import curve_fit
+# import scipy.signal
+from scipy.interpolate import UnivariateSpline
+# from scipy.interpolate import BSpline, SmoothBivariateSpline
+
+# from specutils import SpectralRegion
```
crawfordsm, line 31: Once complete, the imports can be cleaned up.

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+
+    zeropoint : float, optional
+        Default is set to 48.60. Conversion factor for converting mag into flux.
+
+    Returns
+    -------
+    Corresponding quantity in flux units.
+    """
+
+    c = 2.99792458e18  # speed of light, in Angstrom/second
```
crawfordsm, line 60: Use astropy.constants

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+    Generate the sensitivity function (or sensfunc) needed for the flux calibration of
+    a 1D spectrum.
+
+    Parameters
+    ----------
+    obj_wave : a 1D numpy array
+        The 1D wavelength array of the spectrum to be flux calibrated.
+
+    obj_flux : a 1D numpy array
+        The 1D flux array of the spectrum to be flux calibrated.
```
crawfordsm, line 76: Something to discuss, but instead of an wave and flux here, I think we want to have users input a Spectrum1D object to the function.

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+
+    std : str
+        Name of the standard star file in the specreduce/datasets/onedstds
+        directory to be used for the flux calibration. The user must provide the
+        subdirectory and file name. For example:
+
+        >>> genfluxcal(obj_wave, obj_flux, std='/ctiocal/eg21', mode='poly', polyorder=6)  \
+        # doctest: +SKIP
+
+        If no std is supplied, will raise a ValueError.
```
crawfordsm, line 86: I think it would be my preference to separate the Input of the Standard star and the code.   So I think it would be useful to have it's own function for reading in the supplied standard stars (to be determined) and that to this function, a user should just input a Spectrum1D of the standard star they want to use (that way, it isn't limited to the given list).

specreduce/core.py
crawfordsm: See comment above, but could these also be read in with the specutils functionality?

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+
+    if os.path.isfile(os.path.join(std_dir, std)):
+        std_wave, std_mag, std_wth = np.genfromtxt(
+            os.path.join(std_dir, std), skip_header=1, unpack=True
+        )
+
+        # Convert standard star from magnitude units to flux
+        std_flux = _mag2flux(std_wave, std_mag)
+
+        # Define an array of the three most prominent Balmer lines
```
crawfordsm, line 122: As a design principle, I think it is useful to provide a way for users to masks regions that should be ignored, but this shouldn't hard coded into function.  I'd suggest changing this to a parameter that is passed to the function perhaps using Spectral Regions.   This would make the function more general.

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+        ratio = np.abs(np.array(std_flux_ds, dtype="float") /
+                       np.array(obj_flux_ds, dtype="float"))
+
+        # Interpolate flux calibration (sensfunc) onto object's wavelength grid
+        # 3 modes of interpolation available: 'linear', 'cubic spline', and 'polynomial'
+
+        # To check if a mode of interpolation has been specified
+        if mode not in ("linear", "spline", "poly"):
+            mode = "spline"
+            print("Warning: No interpolation mode has been set"
```
crawfordsm, line 154: Use python internal warning rather than print functions so that users can set the warning levels.

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+            plt.ylabel("Standard flux")
+            plt.legend()
+            plt.show()
+
+        else:
+            sensfunc = np.zeros_like(obj_wave)
+            msg2 = "Warning: No valid standard star file has been entered"
+            print(msg2)
+
+            return 10**sensfunc
```
crawfordsm, line 222: is meant to be outside of the if/else block?

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+            isbalmer = np.where((balmer >= std_wave[i] - std_wth[i] / 2.0) &
+                                (balmer < std_wave[i] + std_wth[i] / 2.0))
+
+            # Check to see if the bin does indeed contain observed spectra,
+            # with no presence of the Balmer lines defined previously
+            if (len(rng[0]) > 1) and (len(isbalmer[0]) == 0):
+                obj_flux_ds.append(np.nanmean(obj_flux[rng]))
+                obj_wave_ds.append(std_wave[i])
+                std_flux_ds.append(std_flux[i])
+
```
crawfordsm, line 142: It might be worthwhile to take a look at the specutils spectral binning and resampling tools and see if they could be used here.

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+    Parameters
+    ----------
+    obj_wave : a 1D numpy array
+        The 1D wavelength array of the spectrum to be flux calibrated.
+
+    obj_flux : a 1D numpy array
+        The 1D flux array of the spectrum to be flux calibrated.
+
+    std : str
```
crawfordsm, line 78: This could use a more descriptive name

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+
+def applyfluxcal(obj_flux, sensfunc):
+    """
+    To apply the sensfunc on the flux of the object as determined with the
+    genfluxcal method.
+
+    Parameters
+    ----------
+    obj_flux : a 1D numpy array
+        The 1D flux array of the spectrum to be flux calibrated.
```
crawfordsm, line 233: This would be useful to work on a Spectrum1D object

specreduce/core.py
```diff
@@ -2,3 +2,246 @@
...
+            plt.legend()
+            plt.show()
+
+        else:
+            sensfunc = np.zeros_like(obj_wave)
+            msg2 = "Warning: No valid standard star file has been entered"
+            print(msg2)
+
+            return 10**sensfunc
+
```
crawfordsm, line 223: It also might be worthwhile to discuss what might be the best object for this to return -- it might make sense to return something with error information so that can be propagated.

specreduce/core.py
kakirastern: Agreed, let me find out how

specreduce/core.py
kakirastern: Yeah, true

specreduce/core.py
kakirastern: Sure, will keep that in mind

specreduce/core.py
kakirastern: Yeah, will use the following instead:

```python
>>> from astropy import constants as const

>>> const.c.to('Angstrom/s')
2.9979246e18 Angstrom/s
```

specreduce/core.py
kakirastern: Let me try this out and compare with other methods...

specreduce/core.py
kakirastern: Yeah, agreed. I can see how [`SpectralRegion`](https://specutils.readthedocs.io/en/stable/spectral_regions.html) objects can be applied in this context now.

specreduce/core.py
kakirastern: Okay sure, I could do something like the following: 

```python
import warnings

warnings.warn('No interpolation mode has been set.')
```

specreduce/core.py
kakirastern: Sounds like a good idea. Will implement this shortly.

specreduce/core.py
kakirastern: Yup, sounds good.

specreduce/core.py
kakirastern: Yeah, that would be better than returning some `10**np.zeros_like()` object which would give the false impression that flux calibration has indeed been successful.

specreduce/core.py
kakirastern: Yeah, @jradavenport. This has been a draft based on your `pydis` source code, except I cleaned it up a bit.

jradavenport: For reference, here's where my version currently is
https://github.com/jradavenport/specreduce/blob/workflow_exploration/specreduce/fluxcal.py
We're definitely doing duplicate work right now, LOL

specreduce/core.py
kakirastern: Yeah, could compare different approaches say with using `astropy.table` or even some similar `specutils` and refactor this code snippet and see which one is more suitable for the context.

specreduce/core.py
kakirastern: Yeah @jradavenport or you could open a new PR and take it from here. I could close this once your version is merged.

specreduce/core.py
kakirastern: Think so, will double check and make changes accordingly.

specreduce/core.py
jradavenport: I agree, but can (should) we support _both_ an array or a Spectrum1D object?

kakirastern: Yeah, we could support both a `numpy` array and `Spectrum1D` object by checking the type with `isinstance()` first. But personally would prefer using `Spectrum1D`.

specreduce/core.py
jradavenport: newbie Q: do we have a demo for how to do this properly?

specreduce/core.py
crawfordsm: If it's an array, you need both a wavelength and a flux.  In the future, you might also need the error array and data quality array or other information.  Either way, this will be additional parameters in the function call (which can be confusing and harder to maintain) and then checking of those parameters (which can be more to maintain and redundant work).   

For these reasons, I would argue passing in the Astropy objections to the functions would be a best practice unless there is another python native way to represent it or no other information is needed.

specreduce/core.py
kakirastern: Or we could collaborate on this PR while keeping it open.

kakirastern:
Will need to change "badlines" to use SpectralRegion as their representations in the `standard_sensfunc` class function.

jradavenport:
this looks quite good, and i like the new class structure.

I find it very valuable (both for development and as a potential user) to have a demo notebook or file. I know there are astropy tutorials, but could/should we include one alongside this? i.e. I don't quite know how the API works in my modeled workflow yet

specreduce/core.py
kakirastern: > this looks quite good, and i like the new class structure.
> 
> I find it very valuable (both for development and as a potential user) to have a demo notebook or file. I know there are astropy tutorials, but could/should we include one alongside this? i.e. I don't quite know how the API works in my modeled workflow yet

We could include examples in the `specreduce` docs. And moreover, I could start working on some tutorial notebooks on Learn Astropy regarding the `fluxcal` class as a start in parallel.

kakirastern:
> this looks quite good, and i like the new class structure.
> 
> I find it very valuable (both for development and as a potential user) to have a demo notebook or file. I know there are astropy tutorials, but could/should we include one alongside this? i.e. I don't quite know how the API works in my modeled workflow yet

@jradavenport We could include examples in the specreduce docs. And moreover, I could start working on some tutorial notebooks on Learn Astropy regarding the fluxcal class as a start in parallel.

kakirastern:
> Will need to change "badlines" to use `SpectralRegion` as their representations in the `standard_sensfunc` class function.

Regarding the `standard_sensfunc` class function: I tried to use the above approach, but realized it may not be practical as some `scipy` magic for interpolation is needed for the interpolation of the calibration. This is such that it would be better for the `badlines`to remain being presented as `numpy` arrays instead of `specutils`'s `SpectralRegion`s. Besides, the widths of the `badlines` should ideally match those of the ones contained in the `standard` table, so that would render the use of `SpectralRegion` difficult since we would need to know ahead of time if such lines are present in the spectrum of the standard star in order to set the lower bound and upper bound of each `badlines` `SpectralRegion` properly. (We might not want the extra overheads incurred anyway IMO.)

By the way, just was thinking maybe we could enable the creation of some blank/empty `SpectralRegion` in the near future for `specutils`. This is because if I use a `for` loop to add all `SpectralRegion`s, I will need some blank object as the starting point/initial state of the `SpectralRegion` for the `badlines` to be added in later, to make code cleaner. Just an idea now anyway.

chris-simpson:
Are there any thoughts about flux calibrating a 2D spectral image?

kakirastern:
> Are there any thoughts about flux calibrating a 2D spectral image?

@chris-simpson Think there should be a follow-up (but separate) PR for dealing with flux calibration of a 2D spectrum.

tepickering:
in the interest of moving things forward, i'm going to go ahead and merge this now. then i will go in and make the changes to mesh in the work i've done on spectrophotometric standard stars.

PR #53, opened 2020-04-09
Operation class

crawfordsm:
This is a pull request with an initial template of the SpecReduce class and an example using boxcar smoothing for the extraction.


jradavenport:
this doesn't include `trace`, should we dump that too (recognizing there are some further ideas about trace that need to be implemented)

eteq:
@jradavenport - as initially formulated the suggestion was made (I forget by who...) to "assume the trace object exists" to provide a minimal example of the `SpecreduceOperation` class, because we're thinking there's need for some sort of data object to hole the trace rather than just sending the raw trace function around.  But now that you say it perhaps it makes sense to include some trace to have working tests of this minimal example...

eteq:
Another thought discussed in slack was to pull in the DRAGONS pipeline trace code as a test of the concept that the operations can be modular, so that might be an alternative trace code to use for the above as an experiment.

tepickering:
in the interest of moving things forward i'm just going to go ahead and merge this and fix the issues in a new PR.

PR #59, opened 2020-04-23
Make classes for atmospheric extinction data

tepickering:
This PR adds some classes to wrap optical and IR extinction data that can be used in the spectral data reduction process. This includes access to the optical extinction curves included in the `specreduce_data` package as well as a near/mid IR atmospheric transmission model.


tepickering:
gonna go ahead and merge this in the interest of keeping the ball rolling and not hold up the other stuff i'm working on.

PR #61, opened 2020-04-24
line_list_utils - query to NIST 

HeloiseS:
- changed `table_utils` to `line_list_utils`
- `query_nist` function added
- tests also created for `line_list_utils`

Note: 
- query_nist will fail if no wavlengths are found uner the observed column of the NIST database
- we are still exploring whether to use the Ritz or observed column. This is just a start.


specreduce/tests/test_line_list_utils.py
```diff
@@ -0,0 +1,20 @@
+import ../line_list_utils as lu
```
tepickering, line 1: lose the / in this import line...

PR #69, opened 2021-04-23
Add jradavenport notebooks

eteq:
This adds the notebooks @jradavenport added as part of the work described in astropy/astropy-project#169 to specreduce in a `notebook_sandbox` directory.  We may or may not want to integrate these with the docs somehow, but I'm thinking it's better to get these *somewhere* in specreduce now and sort out making it a bit cleaner later.

closes astropy/astropy-project#169

cc @tepickering


tepickering:
thank you! agree that it's best to get these in here now as a documented starting point.

PR #73, opened 2021-12-09
Bug fixes in boxcar extraction code

ibusko:
This PR implements bug fixes to specreduce's BoxcarExtract code:

- code that avoids taking a slice outside the image array;
- correct positioning of the sky extraction apertures;
- code that handles fractional pixels
- adds a default units (DN) to the extracted flux array, when units are not present (otherwise it's not accepted as part of a Spectrum1D instance).

With those fixes in place, the class can be used in data like the one in here: https://github.com/ibusko/dat_pyinthesky/blob/spectral_extraction_tests/jdat_notebooks/spectral_extraction_tests/specreduce_extract_jwst.ipynb, where the spectrum positioning and the presence of negative background regions make the bugs very visible. 

A unit test was added as well.

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -80,37 +80,33 @@ class BoxcarExtract(SpecreduceOperation):
             self._extract_from_box(img, i, low_end, high_end, onedspec)
 
             # now do the sky fit
-            # TODO add fractional pixel handling
-            itrace_line = int(trace_line[i])
-            sky_y = np.append(
-                np.arange(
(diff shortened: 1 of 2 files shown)


tepickering:
looks like i needed to click a button to get actions to run for first time contributors. the tests themselves all pass, but there's come codestyle whinging. i personally prefer wider lines, but 100 is the astropy convention. once that's cleaned up, i'll go ahead and merge this. everything else looks good! thanks!

PR #79, opened 2022-01-28
Introduce JWST Boxcar Development Notebook from Ivo Busko

duytnguyendtn:
Migrating @ibusko's notebook exploring implementing the JDAT notebook boxcar extraction algorithm into specreduce: https://github.com/ibusko/dat_pyinthesky/blob/spectral_extraction_tests/jdat_notebooks/spectral_extraction_tests/specreduce_extract_jwst.ipynb

This notebook also contains changes during my technical review

Co-authored-by: Ivo <busko@stsci.edu>
Co-authored-by: ojustino <ojustino@users.noreply.github.com>


duytnguyendtn:
I can't request reviewers, but @tepickering should definitely review the algorithm we have here; also @ojustino should look at this just to rubber stamp the "final" version, since we dont' have @ibusko to confirm the small changes I made

tepickering:
i think the `kosmos` stuff can now be replaced with what's currently implemented in specreduce. at least for the simple, flat traces in that test image. however, it's also fine to leave for now to track the changes between what was done before and what can be done now.

ojustino:
Hi @duytnguyendtn, the notebook runs for me without issues. I agree with @tepickering that it would be good to switch the trace implentation from `kosmos` to `specreduce`, but that might be a task for later since getting the extraction algorithms into PRs seems to be the higher priority for now. 

I have a couple of suggestions on the notebook itself:
- The call to `sys.path.insert()` in cell 1 is no longer necessary, so I would remove it.
- To avoid a warning, the second `ax1.plot()` call in cell 15 should have `':'` as its second argument instead of `'k:'`. `color` is specified in a later argument and having that curve be black doesn't make sense.

duytnguyendtn:
Good catch @ojustino ! I've updated the notebook. 

> leave for now to track the changes between what was done before and what can be done now.
This was my intention with introducing this notebook as it is @tepickering. I imagine that substitution can happen as a follow up to this effort?

tepickering:
that's what i was thinking. once i get the trace fitting implemented, i'll update this notebook as part of that effort.

tepickering:
shall i go forth and merge this?

duytnguyendtn:
@tepickering Yes please! I think we're good on our side to merge!

PR #82, opened 2022-02-16
Boxcar extraction using Trace class

ibusko:
This PR modifies the boxcar extraction code to directly interface with the Trace class in `specreduce`. 

It also removes references to background extraction from the unit test module (`test_extract.py`), and adds a few unit tests to check corner cases involving fractional pixels.

Support for 2D "transposed" spectral images (where the dispersion axis is the first axis), and 3D spectral cubes, was partially added but not fully implemented and tested. Transposed spectral images are not yet supported by the Trace class, and 3D cubes were deemed low priority at this point.

Note that the `tracing.py` module currently does not provide any tool to actually find a trace in a spectral image (although it contains a function to build an image to be used in unit tests). In the original notebook the trace finding function was provided by the `kosmos` package. 

A new notebook was added (`notebook_sandbox/jwst_boxcar/boxcar_extraction.ipynb`) to demonstrate the use of `Trace` and `BoxcarExtract` in practice.

Currently there is no support for a WCS, if present in the data.


specreduce/extract.py
```diff
@@ -3,216 +3,162 @@
...
+            cross-dispersion axis
+
+
+        Returns
+        -------
+        spec : `~specutils.Spectrum1D`
+            The extracted 1d spectrum expressed in DN and pixel units
+        """
+        # this check only applies to FlatTrace instances
+        if hasattr(trace_object, 'trace_pos'):
```
kecnry, line 142: what happens for non-FlatTrace instances?  Do we support those or should we raise an error in an else?

specreduce/extract.py
```diff
@@ -3,216 +3,162 @@
...
+            self.center = trace_object.trace_pos
```
kecnry, line 143: `self.center` isn't currently used anywhere except the following check... do we need to set it?  And should we update the `ValueError` to mention `trace_object.trace_pos` instead since that is the value that the user could actually change?

specreduce/extract.py
```diff
@@ -3,216 +3,162 @@
...
-            ),
-            np.max(
-                trace_line + (self.apwidth + self.skysep + self.skywidth) * 2
-            )
-        )
-
-        return fig
+    # TODO: what is a reasonable default?
+    # TODO: int or float?
+    width: int = 5
```
kecnry, line 116: the `_ap_weight_image` function says it takes `width` as a float, so should we make it a float here as well?  Or do we want to leave it as an integer (as was the case for the previous implementation)?

specreduce/extract.py
```diff
@@ -3,216 +3,162 @@
...
+
+    # TODO: should disp_axis and crossdisp_axis be defined in the Trace object?
```
kecnry, line 118: My personal opinion would be to either put all options (except maybe `image` and `trace_object` either in the class itself or as kwargs, but mixing the two will probably cause confusion.

kecnry (commented):
I'll plan to go through the notebook and some test-cases next week, but just wanted to add a few comments on the `BoxcarExtract` class itself.

specreduce/extract.py
ibusko: They will be silently ignored. This shouldn't be a problem, since `trace_pos` only exists in `FlatTrace` instances. Although this suggests in turn that the test should be moved to inside the `FlatTrace` class itself. I didn't do that because I didn't want to modify the `tracing.py` code.

specreduce/extract.py
ibusko: No, we don't need to set `self.center`. I did it in this way just so the existing code could be reused.

specreduce/extract.py
ibusko: I think it should be a float. The code works in the same way with floats or integers, and float is more general.

specreduce/extract.py
ibusko: I personally think the axis information should go in the `Trace` instance. That however requires changes in `tracing.py` to allow it to handle both cases of 2D spectra.

specreduce/extract.py
tepickering: agreed.

notebook_sandbox/jwst_boxcar/boxcar_extraction.ipynb
```diff
@@ -0,0 +1,403 @@
...
+    "spectrum = boxcar(image, trace)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# this one crashes with \"UnitConversionError: 'pix' and 'Angstrom' (length) are not convertible\"\n",
```
tepickering: this looks like a `Spectrum1D` quirk. the units are correctly set to `u.pixel` where this spectrum is created which is intended because wavelength calibration hasn't happened yet.

specreduce/extract.py
```diff
@@ -3,216 +3,162 @@
...
+            crossdisp_axis,
+            image.shape)
+
+        # extract. Note that, for a cube, this is arbitrarily picking one of the
+        # spatial axis to collapse. This should be handled by the API somehow.
+        ext1d = np.sum(image * wimage, axis=crossdisp_axis)
+
+        # TODO: add uncertainty and mask to spectrum1D object
+        spec = Spectrum1D(spectral_axis=np.arange(len(ext1d)) * u.pixel,
+                          flux=ext1d * getattr(image, 'unit', u.DN))
```
tepickering, line 162: the flux unit here should inherit from the input image, if possible, and only use `u.DN` as a last case fallback. gain corrections and similar calibrations should be applied before this step. in other use cases, there may already be a flux calibration applied to 2D data or the flux may be a count rate.

notebook_sandbox/jwst_boxcar/boxcar_extraction.ipynb
```diff
@@ -0,0 +1,403 @@
...
+    "# print(spectrum.wavelength)"
```
tepickering: ah, turns out this is expected behavior. you want to use `spectrum.spectral_axis` to display it in the original units, `u.pixel`. per the `specutils` docs, `spectrum.wavelength` tries to convert to `u.angstrom` which will not work here without wavelength calibration.

tepickering (changes requested):
a few minor comments and requests. the most substantive one is the hard-coding of units in the output `Spectrum1D`. it will be encouraged to use `CCDData` as the input data format for `specreduce` which will allow for input data to contain flux units and uncertainty arrays. that information should be checked for and utilized if available.

tepickering:
i am not sure why the `devdeps` test is failing. it doesn't look like it's related to anything we're doing. i will investigate further and apply hot fixes to `main` as needed.

specreduce/extract.py
ibusko: I added code to support flux units propagation from the image into the extracted spectrum.

notebook_sandbox/jwst_boxcar/boxcar_extraction.ipynb
ibusko: Good point. I rewrote the comment.

specreduce/extract.py
ibusko: I think adding support for WCS is coming soon in a separate ticket. A properly set spectral WCS instance renders `disp_axis` and `crossdisp_axis` redundant. Maybe we should leave the extraction API as is for now, and work the axis issue when implementing WCS? 
In support to @tepickering 's argument, I think  extractions on "transposed" spectral images was deemed low-priority at some point.

ojustino (commented):
### `extract.py`
1. I think we should either take each of `image`, `trace_object`, and `width` as arguments of `BoxcarExtract`'s `__call__()`/`__init__()` *or* have them all be assigned as class attributes. Having some be set one way and some the other isn't intuitive to me.

I lean toward having them all as arguments of `__call__()` for increased flexibility with the `BoxcarExtract` object and since some other astropy utilities (e.g. `modeling`) also work this way. I'm also writing `HorneExtract` like this. Whichever route we go, consistency and intuitiveness should be our biggest concerns.

2. This is more germane to the upcoming tracing discussion (#83), but I think allowing for interchangeability of dispersion and cross-dispersion axes like we do here is better than assuming each is always on a specific axis. We don't have a mandate to write a full pipeline, so in my mind, that limits the number of assumptions we can make about the spectra we receive.

My ideal would be to store dispersion axis info in trace objects, which would make those arguments unneeded in `BoxcarExtract`. Barring that, I think the current approach here is OK.

3. What do we think of having `_get_boxcar_weights()` and `_ap_weight_images()` as class methods of `BoxcarExtract` instead of free-floating functions?

4. The previous iteration of `BoxcarExtract` had some plotting functionality. Does it make sense to include a rudimentary version of that here?

5. Some lines run long on characters and could be separated.

### `test_extract.py`
Would it be useful to include real data as part of our tests? I can share several 2D VLT spectra alongside their pipeline's 1D extractions if so.

### `boxcar_extraction.ipynb`
The notebook runs for me and I find the expository text at the beginning helpful.

1. I think it would be useful to see the pipeline 1D spectrum on the same plot as the extracted 1D spectrum at the end.
2. I suggest rearranging imports into a more standard order (`from`s, then regular `import`s, etc.)

specreduce/extract.py
ojustino: I second Ivo's idea suggestion to perform this check inside the `FlatTrace` class itself and mentioned it in #83. If we don't do that, I agree with Tim's point that `isinstance()` makes this check's intentions more clear.

specreduce/extract.py
tepickering: we need to be a bit careful how deeply we go down the WCS rabbit hole and where. a proper spectral WCS requires a wavelength calibration which may or may not be available at the extraction step. in fact, in a common use case, the aperture used for extraction is then applied to a arc lamp image to extract a spectrum that is then used to determine the wavelength calibration. using WCS information if its there is fine and would be very handy for re-analysis use cases, but we can't assume it will always be there.

tepickering:
looks like the `devdeps` failure resolved itself. it was affecting several other packages in the same way according to the convos i saw on the `astropy` slack...

tepickering:
thanks for all this work, @ibusko, and for the detailed review, @ojustino! per @ojustino's points:

1. i completely agree here. those should be arguments to `__call__()` so that the same `BoxcarExtract` configuration can be reused for multiple images/traces/widths.
2. i'll follow up with my thoughts on this in #83.
3. `_get_boxcar_weights()` should definitely be a class method, i think, since its specifically related to how boxcar extraction is done. `_ap_weight_image()` looks more general purpose, though, and may be applicable to other extraction methods.
4. check plots are nice to have. on the other hand, a `Spectrum1D` object is being returned which has its own built-in `plot()` method. that covers the use case of viewing extracted flux vs pixel position along the trace.
5. the codestyle test that CI runs is configured to use a max line length of 100. i'm open to changing that, but setting it too low can lead to other readability issues, in my opinion...

specreduce/extract.py
kecnry: It looks like all options have now been moved from the class to arguments to `__call__` resolving my original comment and the further discussion about moving axis information to `Trace` is captured in #83.  Are we safe to mark this conversation as resolved for the sake of getting this PR approved?

ibusko: From my viewpoint, it can be marked as resolved.

specreduce/extract.py
kecnry: after discussion with @ibusko and @ojustino - I replaced the `hasattr` with an `isinstance` but left the check in here for now with a comment that we can remove it if/when its migrated to the trace class.

kecnry:
> `_get_boxcar_weights()` should definitely be a class method, i think, since its specifically related to how boxcar extraction is done. `_ap_weight_image()` looks more general purpose, though, and may be applicable to other extraction methods.

`_ap_weight_image` calls `_get_boxcar_weights` so for now I moved them nested into the `__call__`.  If we need them elsewhere down the road, we can move them out or to class methods.

I think this is now ready for re-(final?)-review (assuming CI passes)

PatrickOgle:
Tried out boxcar_extraction.ipynb:
1) Basic functionality is there-input image, trace, width and get out spectrum
2) Output flux is a factor of exactly 10^6 smaller than the comparison x1d product.  Not sure why...maybe a function of post-extraction flux calibration applied to the x1d product.
3) Basic extractions should be unitless, anyway--counts vs. pixel.  Wavelength and flux calibration get applied later.

kecnry:
> Output flux is a factor of exactly 10^6 smaller than the comparison x1d product. Not sure why...maybe a function of post-extraction flux calibration applied to the x1d product.

That's a good point - we did pull the following out from the code when migrating from the notebook.  I'm not sure where this factor best belongs (somewhere in the code without the factor being hardcoded for JWST or maybe in the `boxcar_extraction.ipynb` notebook)?

```
# convert from MJy/sr to Jy
ext1d *= 1e6 * jdatamodel.meta.photometry.pixelarea_steradians
```

tepickering (approved):

tepickering:
CI has passed and i think we're in agreement on the changes that have been made. as far as i'm concerned, this is good to merge. there's more work to do, but it depends on tracing functionality being fleshed out further which is work-in-progress.

ojustino (approved):
I agree and also approve -- looks like we've addressed everything that was immediately crucial.

kecnry:
Also happy to merge, thanks everyone!

Issue #83, opened 2022-02-23
Spectral tracing discussion


ojustino:
This issue is meant as a forum to discuss the new tracing classes in `specreduce/tracing.py` since the PR that introduced them has already been merged.

I'll start by saying all of `Trace`, `FlatTrace`, and `ArrayTrace` work smoothly for me.

Here are some avenues of further discussion and possible changes to consider:

- **We don't currently have a way to calculate a trace based on an image.** I support creating one, but is it a priority? If so, I think @jradavenport had said that we were OK to port code from the `kosmos` `trace()` function into specreduce.
- As @ibusko mentioned [here](https://github.com/astropy/specreduce/pull/82#discussion_r810504358), it would be nice to validate that the trace position is greater than 1 and less than the cross-dispersion axis' length **as the `Trace` object is being created** instead of at the beginning of the extraction process.
- In response to @tepickering's comment [here](https://github.com/astropy/specreduce/pull/82#discussion_r811454384), I support **including dispersion/cross-dispersion axis information as attributes of `Trace` objects** instead of as arguments of an extraction class' `__call__()` method. If we want to allow for more interchangeability, perhaps they could be set as arguments of `Trace` objects. If not, we could hard-code our assumption that axis 1 is dispersion and axis 0 is cross-dispersion into `Trace`. Either way, extraction classes would get that information for free since they must take a `Trace` object as an argument.
- **Should all `Trace` objects contain an `image` attribute?** `ArrayTrace` and `FlatTrace` don't need an image, but future `Trace` implementations may. Including the image in the `Trace` object would also decrease the number of arguments needed for extraction objects. However, @ibusko brought up a point that you may not always want to perform an extraction on the exact image from which you created a trace.

kecnry:
> Should all Trace objects contain an image attribute?

My current thinking is to not take the image unless it is required for the trace.  If we want to implement detecting a trace from an image, I'd propose implementing `trace = ArrayTrace.from_image(image)` instead.  This then makes it clear to the user, I think, when the image is actually being used and when it is not.

tepickering:
object finding and calculating traces based on detected objects is stuff i am actively working on, but not quite ready for a PR. i agree with adding an attribute to `Trace` objects to define which axis is being traced along. that will commonly be the dispersion axis, but not always! the signal finding/trace fitting tools could be used to trace emission lines along a spatial axis, for example. 

[...]

PR #85, opened 2022-03-04
Added kosmos trace algorithm

ojustino:
I ported a trace algorithm from the `KOSMOS` repository's [apextract.py](https://github.com/jradavenport/kosmos/blob/fca3b87f59825d8078f13bbbcc03b6c782ad3c21/kosmos/apextract.py). I refactored much of it and took on one of wishlist requests of using `astropy.modeling`'s `Gaussian1D` instead of a custom function to create Gaussians. The fits end up being slightly different, but I have a local notebook where I've verified that the calculated inputs to the respective fitters are the same in this version and the original.

I wrote it as a descendant of `Trace` instead of as a class method of `ArrayTrace` as suggested in the [tracing discussion](https://github.com/astropy/specreduce/issues/83#issuecomment-1049078655). `ArrayTrace` requires an initial array, while the `KOSMOS` implementation can run with no guess or just a float. I didn't think requiring an array when it wasn't needed made sense, though we can change the format if I've missed something.

The original has arguments for specifying wavelength and spatial directions that I commented out since it seems that's still up for discussion. I also left out functionality for displaying the trace after the calculation.

### DIFF
--- a/specreduce/tracing.py
+++ b/specreduce/tracing.py
@@ -193,8 +193,9 @@ class KosmosTrace(Trace):
         if (self.window is not None
             and (self.window > img.shape[self._disp_axis]
                  or self.window < 1)):
-            raise ValueError(f"window must be >= 2 and less than {self.image.shape[self.disp_axis]} "
-                             "the length of the image's spatial direction")
+            raise ValueError(f"window must be >= 2 and less than "
+                             "{self.image.shape[self.disp_axis]} the length of the "
+                             "image's spatial direction")
         elif self.window is not None and not isinstance(self.window, int):
             warnings.warn('TRACE: Converting window to int')
             self.window = int(self.window)
--- a/specreduce/tracing.py
+++ b/specreduce/tracing.py
@@ -1,11 +1,14 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 from dataclasses import dataclass
+import warnings
 
-import numpy as np
+from astropy.modeling import CompoundModel, fitting, models
 from astropy.nddata import CCDData
+from scipy.interpolate import UnivariateSpline
+import numpy as np
 
-__all__ = ['Trace', 'FlatTrace', 'ArrayTrace']
+__all__ = ['Trace', 'FlatTrace', 'ArrayTrace', 'KosmosTrace']
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+    """
+    Trace the spectrum aperture in an image.
+
+    Chops image up in bins along the dispersion (wavelength) direction,
+    fits a Gaussian within each bin to determine the trace's spatial
+    center. Finally, draws a cubic spline through the bins to up-sample
+    trace along every pixel in the dispersion direction.
+
+    (The original version of this algorithm is sourced from James
+     Davenport's `kosmos` repository.)
```
kecnry, line 129: would it be useful to provide a link?  Are there any license implications?

specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+
+    Parameters
+    ----------
+    image : CCDData or array-like, required
+        The image over which to run the trace. Assumes cross-dispersion
+        (spatial) direction is axis 0 and dispersion (wavelength)
+        direction is axis 1.
```
kecnry, line 136: but it has a `disp_axis` attribute.  If there is somewhere that requires this assumption, then maybe we should just raise a `ValueError` if its set to something other than 1?

specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+        divide the image. Use fewer if KosmosTrace is having
+        difficulty, such as with faint targets. Minimum bin size is 4.
+        [default: 20]
+    guess : int, optional
+        A guess at the trace's location in the cross-dispersion
+        (spatial) direction. If set, overrides the normal max peak
+        finder. Good for tracing a fainter source if multiple traces
+        are present. [default: None]
+    window : int, optional
+        Fit the trace to a region with size `window * 2` aroudn the
```
kecnry, line 148:
```suggestion
        Fit the trace to a region with size `window * 2` around the
```

specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+
+        # guess the peak width as the FWHM, roughly converted to gaussian sigma
+        yy = np.arange(len(ztot))
+        yy_above_half_max = np.sum(ztot > (np.nanmax(ztot) / 2))
+        width_guess = yy_above_half_max / 2.355
+
+        # enforce some (maybe sensible?) rules about trace peak width
+        width_guess = (2 if width_guess < 2
+                       else 25 if width_guess > 25
+                       else width_guess)
```
kecnry, line 198: this scares me a bit... should we at least add warnings?

specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+                    y_bins[i] = popt_tot.mean_0.value
+                else:
+                    y_bins[i] = popt_i.mean_0.value
+                    popt_tot = popt_i
+
+            except RuntimeError:
+                # NOTE: would this happen? if not, pguess and try/except are unneeded
+                warnings.warn(f"TRACE: Fitting bin {i} caused RuntimeError; "
+                              "reverting to initial guess")
+                popt_i = pguess
```
kecnry, line 257: if we're not sure if this is needed, I would vote to remove it.  If we are keeping it in, then maybe require passing a flag to fallback on the default?  Otherwise I'm worried it could result in unexpected results (especially if anyone has warnings suppressed).

kecnry (commented):
I would still vote in favor of `trace = ArrayTrace.kosmos_from_image(image, bins, guess, window)` or `trace = ArrayTrace.from_image(image, alg='kosmos', **kwargs)` (which wouldn't require an array, it _creates_ and sets the array) **unless** there will be additional methods we expect to add for this that won't be applicable to the general parent class (or if we expect that to be the case for other algorithms).  My thinking is that the inheritance is not as obvious to the user that they really have the same functionality as an `ArrayTrace` object, just constructed differently.  But we can defer that decision for a while, if necessary, as it really is just moving around existing code.

This also really needs a test case, which will also make it easier to make sure its behaving as expected and thoroughly step through the code.

specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+    disp_axis = 1
+    
+    def __post_init__(self,
+                      # Saxis=0, Waxis=1, display=False
+                      ):
+        if self.bins < 4:
+            raise ValueError('bins must be >= 4')
+        if not isinstance(self.bins, int):
+            warnings.warn('TRACE: Converting bins to int')
+            self.bins = int(self.bins)
```
duytnguyendtn, line 174: I would switch these two so that the bins enforcement happens on rectified data
``` suggestion
        if not isinstance(self.bins, int):
            warnings.warn('TRACE: Converting bins to int')
            self.bins = int(self.bins)
        if self.bins < 4:
            raise ValueError('bins must be >= 4')
```

specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+
+        if (self.window is not None
+            and (self.window > self.image.shape[self.disp_axis]
+                 or self.window < 1)):
+            raise ValueError("window must be >= 2 and less than the length of "
+                             "the image's spatial direction")
```
duytnguyendtn, line 180: Maybe add feedback to the explicit range?
``` suggestion
            raise ValueError(f"window must be >= 2 and less than {self.image.shape[self.disp_axis]} "
                             "the length of the image's spatial direction")
```

specreduce/tracing.py
duytnguyendtn, line 198: While I'm all for intelligent software, I share @kecnry's concern about these hardcoded values. I'm wondering if these values can be extracted from some attribute of the image that's coming in? But that's outside my expertise

specreduce/tracing.py
```diff
@@ -110,3 +113,164 @@ def __post_init__(self):
...
+            # repeat earlier steps to create gaussian fit for each bin
+            z_i = np.nansum(self.image[ilum2, x_bins[i]:x_bins[i+1]],
+                            axis=self.disp_axis)
+            peak_y_i = ilum2[np.nanargmax(z_i)]
+
+            yy_i_above_half_max = np.sum(z_i > (np.nanmax(z_i) / 2))
+            width_guess_i = yy_i_above_half_max / 2.355
+            width_guess_i = (2 if width_guess_i < 2
+                             else 25 if width_guess_i > 25
+                             else width_guess_i)
```
duytnguyendtn, line 228: See above
> While I'm all for intelligent software, I share @kecnry's concern about these hardcoded values. I'm wondering if these values can be extracted from some attribute of the image that's coming in? But that's outside my expertise

duytnguyendtn (changes requested):
Substituting the `FlatTrace` with `KosmosTrace` in the boxcar extract notebook produced results that by-eye looked appropriate. I strongly suggest a scientific review from someone who can scientifically validate the results (see @PatrickOgle and of course @tepickering)

specreduce/tracing.py
ojustino: I copied these numbers as they were, so maybe we should learn where they come from and whether they're worth keeping before adding warnings. @jradavenport, do you remember how you came up with this range of valid guesses for the width of the trace's peak?

ojustino: I agree on the if/else clause, but the 2.355 in the first line you've highlighted comes from [this equation](https://en.wikipedia.org/wiki/Full_width_at_half_maximum#Normal_distribution) relating full width at half-maximum to standard deviation.

specreduce/tracing.py
ojustino: I think this is another case where understanding the original motivation might help.  @jradavenport, do you remember why you had to account for `RuntimeError`s in the loop for creating Gaussian fits to each bin?

specreduce/tracing.py
tepickering: should use `astropy.stats.gaussian_fwhm_to_sigma` here in lieu of 2.355 for both precision and clarity.

specreduce/tracing.py
ojustino: I'm agnostic on a link, but the original code is MIT-licensed, meaning we can modify and distribute the it without limitation. I believe the only thing we may be obligated to do is include a copy of `KOSMOS`' license here. We could add a note that it only applies to the specific file(s) that contain original or modified code from there.

We also did get [written permission](https://github.com/astropy/specreduce/issues/83#issuecomment-1050096916) from James earlier.

specreduce/tracing.py
ojustino: The class does assume that `disp_axis` is 1, but `disp_axis` is not an argument and our prescribed workflow doesn't include modifying it. What do you think about communicating that more implicitly by calling it `_disp_axis` versus something more explicit like you suggested?

PatrickOgle:
It seems to work on the example data in compare_extractions, replacing FlatTrace with KosmosTrace(img).  However, I do not understand the output format of the trace.  I would expect a 1D array of values with the trace location for each x value in the input image.  Please add a notebook example of how KosmosTrace works and overplot the location of the computed trace on the input 2D image.  Also, the fake dataset  in that notebook would be more interesting if it had a curved trace.

specreduce/tracing.py
kecnry: I think we should either do that or at least ensure its 1 and raise an error otherwise (if we expect to remove this assumption down the road, that might be nice)

specreduce/tracing.py
jradavenport: they were good values I arrived at experimentally to ensure the trace didn't go off the rails... 2 pixels being a minimum to (hopefully) ensure I wasn't locking onto a cosmic ray. 25 pixel maximum to keep me looking at a point source instead of just noise

specreduce/tracing.py
jradavenport: I _believe_ this was to deal with `curve_fit` failing... not clear to me if its needed still

specreduce/tracing.py
ojustino: What size images were you working with? Maybe we can turn the 25 into a fraction of the image's size in the cross-dispersion direction.

ojustino:
I'm collecting current feedback in one comment. Some are noted as still being in progress, but please offer opinions and highlight those that are preventing a merge.

1. **Format.** Is there any more feedback on whether we prefer this functionality to be a method of `ArrayTrace` or inherit from `Trace` as its own class?
2. **Tests.** I'm writing some.
3. **Notebook example?** Patrick has asked for a notebook example, but would one be compelling? Not as much goes into preparing for a trace as with the extractions.
4. **Magic numbers.** Being discussed in [this thread](https://github.com/astropy/specreduce/pull/85#discussion_r822995310).
5. **Licensing.** I'll copy the steps I mentioned [here](https://github.com/astropy/specreduce/pull/85#discussion_r823139423) if there are no objections.

PatrickOgle:
OK, knowing that I can access the trace via  result.trace enabled me to overplot the trace on the image.  Here is an example with bins=5
kosmos_trace_8bins

tepickering:
great! i'll give things a full review and do a notebook run-through myself later today so we can get this merged.

kecnry (approved):
All my comments have either been addressed or postponed to future work.  The code itself and tests look good.  Whichever get merged first (this or #86), it would be useful to rebase the other and update the notebook to cover the case of an initial rough trace -> background subtraction -> kosmos trace -> extraction.

specreduce/tracing.py
```diff
@@ -110,3 +114,177 @@ def __post_init__(self):
...
+
+        if self._disp_axis != 1:
+            raise ValueError('dispersion axis must equal 1')
+
+        if not isinstance(self.bins, int):
+            warnings.warn('TRACE: Converting bins to int')
+            self.bins = int(self.bins)
+
+        if self.bins < 4:
+            raise ValueError('bins must be >= 4')
```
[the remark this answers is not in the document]
tepickering: the useful upper limit is the length of `_disp_axis`. if i'm correctly reading the logic of where `x_bins` is being set, it might actually be `_disp_axis - 1`. if you don't want to add the check now, we can make an issue and tackle it later.

specreduce/tracing.py
tepickering: i think this works more or less well enough for now. DRAGONS has some more sophisticated algorithms for peak finding, centering, and width measurements that we'll likely want to port over. we can compare/contrast methods when we get to that point and adjust as necessary.

tepickering (approved):
left a couple comments. it might be worth adding an upper limit check for `bins`, but it's also fine to leave it and make an issue for it. there's still more work to do on tracing and that fix can be done as part of that.

specreduce/tracing.py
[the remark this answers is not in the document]
ojustino: I added this to the latest commit.

PR #86, opened 2022-03-16
Basic (boxcar) background subtraction

kecnry:
This PR implements `specreduce.background.Background(image, trace/center, separation, width)` support. 

The current API supports the following pseudo-code (see the edited notebook for full details):
```
# simplest case to create a background around a row
bg = Background.two_sided(image, ext_center, separation=bkg_sep, width=bkg_width)

# or by defining and passing the spectrum trace
trace = FlatTrace(image, ext_center)
bg = Background.two_sided(image, trace, separation=bkg_sep, width=bkg_width)

# or to only have the background from above/below
bg = Background.one_sided(image, trace, separation=bkg_sep, width=bkg_width)

# or most generally by passing a list of traces
bg = Background(image, [trace-bkg_sep, trace+bkg_sep], width=bgk_width)

# could pass image-bg to automated trace-detection algorithms to refine further

boxcar = BoxcarExtract()
spectrum = boxcar(image-bg, trace, width=ext_width)
```

In doing so, this:
* exposes some internal methods from Boxcar extract to re-use here
* adds and uses +/- offset support within traces


PatrickOgle:
Please add the following options, which I consider to be essential:
1) Option to use median rather than average to compute the background
2) Option for 1-sided background extraction.

camipacifici:
Very nice!
Question:
If I compute the background like this
`bkg = Background.two_sided(data, trace, separation=7, width=5)`
and I plot it with
`plt.imshow(bkg.bkg_image(data), origin='lower')`
I get
Screen Shot 2022-03-17 at 3 33 47 PM

If instead I use the median
`bkg = Background.two_sided(data, trace, separation=7, width=5, statistic='median')`
and I plot it in the same way, I get
Screen Shot 2022-03-17 at 3 34 55 PM

What am I missing?

camipacifici:
Typo in the description of the PR:
bg = Background.two_side(image, trace, separation=bkg_sep, width=bkg_width)
two_side -> two_sided

kecnry:
@camipacifici 

> If instead I use the median `bkg = Background.two_sided(data, trace, separation=7, width=5, statistic='median')`...

this should be fixed now.  The slicing was collapsing the axes so returning a single value instead of an array, which oddly didn't cause the rest of the code to complain.  Now it returns the same shape for the background_image and it seems reasonable on the example data in the notebook.  Thanks for noticing that!

specreduce/extract.py
```diff
@@ -13,6 +13,66 @@
 __all__ = ['BoxcarExtract']
 
 
+def _get_boxcar_weights(center, hwidth, npix):
```
tepickering, line 16: if these methods are going to be used outside of this file, they should be renamed to remove the leading `_` and added to `__all__`.

kecnry: these are methods we need in both extract and background, but probably shouldn't be user-facing.  Do you still want me to do this and we just won't "advertise" them in the docs?  Or is there some better way to handle this type of situation?

tepickering: if they are going to only used within `specreduce`, then they're fine as-is. my personal style is to keep things private within files, but the official python guidelines treat privacy at the module level. the leading `_` and exclusion from `__all__` should be sufficient to keep them from being advertised.

specreduce/background.py
```diff
@@ -0,0 +1,205 @@
...
+    disp_axis : int
+        dispersion axis
+    crossdisp_axis : int
+        cross-dispersion axis
+    """
+    # required so numpy won't call __rsub__ on individual elements
+    # https://stackoverflow.com/a/58409215
+    __array_ufunc__ = None
+
+    image: CCDData
```
tepickering, line 41: should be `NDData` to match the docstring. or vice versa.

specreduce/background.py
```diff
@@ -0,0 +1,205 @@
...
+    traces: list = field(default_factory=list)
+    width: float = 5
+    statistic: str = 'average'
+    disp_axis: int = 1
+    crossdisp_axis: int = 0
```
tepickering, line 46: only `disp_axis` is needed as an input. `crossdisp_axis` can be defined based on that within `__post_init__`.

specreduce/background.py
```diff
@@ -0,0 +1,205 @@
...
+            to extract the background.
+
+        Returns
+        -------
+        array with same shape as ``image``
+        """
+        if image is None:
+            image = self.image
+
+        return image - self.bkg_image(image)
```
tepickering, line 199: this works for normal `ndarray` instances, but does not work for `NDData` and subclasses. they implement their own arithmetic methods so they can support different ways of handling masks and propagating uncertainties. see https://docs.astropy.org/en/stable/nddata/mixins/ndarithmetic.html for details.

tepickering:
some tests will need to be added before this can be merged.

specreduce/background.py
kecnry: I agree, but see that the other recent classes also define `crossdisp_axis` and some of this logic may move up to the trace objects in the future. Do you have a preference for whether to do that just in this PR or across all classes at once (or maybe inheriting from a base class/mixin that handles that)?

tepickering: technically, if you're accepting both as inputs, there should be validation to make sure they're not equal and that both are in range of the images dimensions. ultimately, `disp_axis` should be an attribute of `Trace` classes and flow down from there as described in #89. so i guess it's fine to leave as-is for now and do all the refactoring then.

kecnry: I like that plan.  I agree that it should be changed, but that #89 would be the best opportunity to handle it consistently across all recent work.

tepickering: i have a proto-PR in the works to cover that and more, but won't be ready in time for next week's release.

specreduce/tests/test_background.py
```diff
@@ -0,0 +1,43 @@
...
+from specreduce.tracing import FlatTrace
+
+
+# Test image is comprised of 30 rows with 10 columns each. Row content
+# is row index itself. This makes it easy to predict what should be the
+# value extracted from a region centered at any arbitrary Y position.
+image = np.ones(shape=(30, 10))
+for j in range(image.shape[0]):
+    image[j, ::] *= j
+image = CCDData(image, unit=u.Jy)
```
ojustino, line 16: A note here that this is the same test image as in `specreduce/tests/test_extract.py` might be helpful for ease of future reading.

specreduce/tracing.py
```diff
@@ -55,6 +56,20 @@ def _bound_trace(self):
...
+        """
+        copy = deepcopy(self)
+        copy.shift(delta)
+        return copy
+
+    def __sub__(self, delta):
+        """
+        Return a copy of the trace shifted by delta pixels perpendicular to the axis being traced
+        """
+        return self.__add__(-delta)
```
ojustino, line 71: It may be helpful if these docstrings noted the (opposite) directions on which these methods act, as well as the axis.

ojustino (approved):
This works without issue for me. I see improved performance as the given trace comes closer to the image's actual trace, which is sensible. The test passes.

I think I missed the conversation on the one and two-sided options, but the implementation looks OK to me. I left a couple of more minor comments that shouldn't prevent a merge but could be helpful for the future.

Based on the commented cell in `notebook_sandbox/jwst_boxcar/boxcar_extraction.ipynb`, it looks like a merge should wait until #85 is finished so the notebook can be updated first.

kecnry:
Rebased on top of #85 and added a call to `KosmosTrace` (using the subtracted image from a `FlatTrace`) into the sandbox notebook.

tepickering:
great! i'll merge once CI completes, if that's ok.

PR #91, opened 2022-03-26
Release prep

tepickering:
This PR fixes an issue in pushing packages to PyPI with the default versions that `setuptools_scm` creates. Tags will need to be created for packages to be published. If the tag contains `test`, it will publish to the Test PyPI index. Otherwise it will publish to the main one. We want to reserve the latter for official releases.

A `CHANGES.txt` file is also added here and contains a basic list of what has been added recently. Feel free to edit and add more details.


pllim:
This has been superseded by #92

PR #92, opened 2022-03-29
MNT: Fix publish workflow

pllim:
This uses https://github.com/OpenAstronomy/github-actions-workflows/blob/main/.github/workflows/publish_pure_python.yml as @astrofrog suggested.

Close #91


.github/workflows/publish-to-pypi.yml
```diff
@@ -1,47 +1,17 @@
...
-
-    - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        user: __token__
-        password: ${{ secrets.PYPI_API_TOKEN }}
+  publish:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    if: ((github.event_name == 'push' && startsWith(github.ref, 'refs/tags')) || contains(github.event.pull_request.labels.*.name, 'Build wheels'))
```
astrofrog, line 12: You could actually remove this, it will then always test out building the wheels and just not publish them except on tags. For the core package I think it doesn't make sense to burn lots of CI time for this but for smaller packages I think it's fine to just let it always run, it's a good way to catch issues before releases.

pllim: OK!

tepickering:
the `doc_test` failure was some internal glitch. i re-ran it and it worked fine.

tepickering (approved):
thanks for this! guess i should have looked at the openastronomy workflows first. i'll edit `CHANGES.txt` to remove the "unreleased" bit before i merge.

pllim:
The OpenAstronomy workflows are very new. I forgot myself until @astrofrog reminded me.

PR #97, opened 2022-03-29
Included KosmosTrace in optimal_extract_VLT.ipynb

ojustino:
This is a follow-up discussed in #85 for after #86 was merged.


tepickering (approved):
looks good! will merge once the CI passes...

PR #105, opened 2022-05-02
Made non-required Horne arguments truly optional

### LINKED ISSUE #104: HorneExtract: mask and unit as optional for array-like image
Currently the docs say they are optional, but there is a check that raises an error.  The `if mask is None` check will also need to happen before comparing `image.shape` to `mask.shape`.

_Originally posted by @kecnry in https://github.com/astropy/specreduce/pull/84#discussion_r861257314_

ojustino:
Should resolve #104.

It was decided that `mask` and `unit` should be optional arguments of `HorneExtract()` when the image is a `numpy` array, but the argument validation routine treated them as requirements. These changes are meant to settle the discrepancy.

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -238,19 +238,19 @@ class HorneExtract(SpecreduceOperation):
                 raise ValueError('image NDData object lacks uncertainty')
 
         else:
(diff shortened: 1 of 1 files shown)


tepickering:
lgtm!

Issue #96, opened 2022-03-29
TST: Set up coverage test


pllim:
I think it is good to have, especially given this library will eventually be used in a lot of science.

Issue #95, opened 2022-03-29
TST: Remove custom ci skip check in Actions


pllim:
Actions now natively supports `[ci skip]` directive (for better or worse), so there is no need for this:

https://github.com/astropy/specreduce/blob/85af34074dcbf034220d044b7691f427404563d4/.github/workflows/tox-tests.yml#L14

Issue #94, opened 2022-03-29
TST: Replace doc_test job with RTD PR builder


pllim:
https://github.com/astropy/specreduce/blob/85af34074dcbf034220d044b7691f427404563d4/.github/workflows/tox-tests.yml#L52-L53

The comment about checking link is inaccurate because it does not run linkcheck defined in `tox.ini`. I see you already have RTD set up, so you should just do this: https://docs.readthedocs.io/en/stable/pull-requests.html

As for linkcheck, it is best run as a cron job. Links can be flaky, there is no need to check it every commit.

tepickering:
the RTD config itself was out of date and broken. addressed in #107 and the suggested configuration change has been made. looks like it's building fine now: https://specreduce--107.org.readthedocs.build/en/107/

PR #107, opened 2022-05-16
Testing updates and post-release fixes

### LINKED ISSUE #93: TST: Handle test warnings
At the time of posting, there are 4 warnings. These should be handled accordingly. You might want to turn all unhandled warnings into exceptions going forward in pytest settings to ensure no surprise warnings creep back in.

You can ignore this one: sunpy/ndcube#514
```
.../ndcube/version.py:27
  .../ndcube/version.py:27: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    for j, piece in enumerate(LooseVersion(version).version[:3]):
```

Is this good or bad? What is `specreduce_data`? Does it block release?
```
.../specreduce/calibration_data.py:34
  .../specreduce/calibration_data.py:34: AstropyUserWarning: Can't import specreduce_data package. Falling back to downloading data...
    warnings.warn(
```

This is not a blocker but assumption is dangerous. Should update the code so it does not generate this warning.
```
specreduce/tests/test_extinction.py::test_missing_extinction_unit
  .../specreduce/calibration_data.py:292: AstropyUserWarning: Input extinction is not a Quanitity. Assuming it is given in magnitudes...
    warnings.warn(
```

Is this warning expected? If so, use `pytest.warns` in the test. If not, fix the test case.
```
specreduce/tests/test_extinction.py::test_transmission_model
  .../astropy/units/function/logarithmic.py:46: RuntimeWarning: divide by zero encountered in log10
    return dex.to(self._function_unit, np.log10(x))
```

tepickering:
This PR updates the CI and testing configuration:

- The main CI workflow is significantly cleaned up and versions of referenced actions are updated.
- A cron workflow is added to monitor upstream development releases and check links in documentation.
- `tox.ini` is updated to support Python 3.10 and Python 3.10 tests are added to the CI.
- Add test to check against oldest supported dependencies.
- Add test against `astropy` release candidates (this may be temporary).

Closes #93 #94 #95 #96

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,7 +70,9 @@ doctest_plus = enabled
 text_file_format = rst
 addopts = --doctest-rst
 filterwarnings =
+    error
     ignore:Can\'t import specreduce_data package:
+    ignore:numpy.ndarray size changed:
 
 [coverage:run]
 omit =
--- a/setup.cfg
+++ b/setup.cfg
@@ -69,6 +69,8 @@ astropy_header = true
 doctest_plus = enabled
 text_file_format = rst
 addopts = --doctest-rst
+filterwarnings =
+    ignore:Can\'t import specreduce_data package:
 
 [coverage:run]
 omit =
--- a/specreduce/tests/test_extinction.py
+++ b/specreduce/tests/test_extinction.py
@@ -1,6 +1,9 @@
+import pytest
+
 import numpy as np
 
 import astropy.units as u
+from astropy.utils.exceptions import AstropyUserWarning
 
 from ..calibration_data import (
     AtmosphericExtinction,
@@ -56,14 +59,16 @@ def test_missing_extinction_unit():
     """
     Test creation of custom model from Quantity arrays
     """
-    wave = np.linspace(0.3, 2.0, 50)
-    extinction = 1. / wave
-    ext = AtmosphericExtinction(extinction=extinction, spectral_axis=wave * u.um)
-    assert(len(ext.extinction_mag) > 0)
-    assert(len(ext.transmission) > 0)
+    with pytest.warns(AstropyUserWarning):
+        wave = np.linspace(0.3, 2.0, 50)
+        extinction = 1. / wave
+        ext = AtmosphericExtinction(extinction=extinction, spectral_axis=wave * u.um)
+        assert(len(ext.extinction_mag) > 0)
+        assert(len(ext.transmission) > 0)
 
 
 def test_transmission_model():
-    ext = AtmosphericTransmission()
-    assert(len(ext.extinction_mag) > 0)
-    assert(len(ext.transmission) > 0)
+    with pytest.warns(RuntimeWarning):
+        ext = AtmosphericTransmission()
+        assert(len(ext.extinction_mag) > 0)
+        assert(len(ext.transmission) > 0)
--- a/specreduce/tests/test_extinction.py
+++ b/specreduce/tests/test_extinction.py
@@ -18,8 +18,8 @@ def test_supported_models():
     """
     for model in SUPPORTED_EXTINCTION_MODELS:
         ext = AtmosphericExtinction(model=model)
-        assert(len(ext.extinction_mag) > 0)
-        assert(len(ext.transmission) > 0)
+        assert len(ext.extinction_mag) > 0
+        assert len(ext.transmission) > 0
 
 
 def test_custom_mag_model():
@@ -29,8 +29,8 @@ def test_custom_mag_model():
     wave = np.linspace(0.3, 2.0, 50)
     extinction = u.Magnitude(1. / wave, u.MagUnit(u.dimensionless_unscaled))
     ext = AtmosphericExtinction(extinction=extinction, spectral_axis=wave * u.um)
-    assert(len(ext.extinction_mag) > 0)
-    assert(len(ext.transmission) > 0)
+    assert len(ext.extinction_mag) > 0
+    assert len(ext.transmission) > 0
 
 
 def test_custom_raw_mag_model():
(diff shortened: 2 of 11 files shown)


pllim:
Awesome. Thanks!

setup.cfg
```diff
@@ -69,6 +69,8 @@ astropy_header = true
 doctest_plus = enabled
 text_file_format = rst
 addopts = --doctest-rst
+filterwarnings =
+    ignore:Can\'t import specreduce_data package:
```
pllim, line 73: If you want to prevent unhandled warnings from creeping back in...

```suggestion
filterwarnings =
    error
    ignore:Can\'t import specreduce_data package:
```

specreduce/tests/test_extinction.py
```diff
@@ -56,14 +59,16 @@ def test_missing_extinction_unit():
     """
     Test creation of custom model from Quantity arrays
     """
-    wave = np.linspace(0.3, 2.0, 50)
-    extinction = 1. / wave
-    ext = AtmosphericExtinction(extinction=extinction, spectral_axis=wave * u.um)
-    assert(len(ext.extinction_mag) > 0)
-    assert(len(ext.transmission) > 0)
+    with pytest.warns(AstropyUserWarning):
```
pllim, line 62: Ideally, you only want the context manager covering the line that actually emits the warning.

setup.cfg
tepickering: when i add the `error` bit, i get tonnes of failures due to `numpy.ndarray size changed` that are coming from upstream. happens both locally via `pytest` and via `tox` which installs everything fresh via `pip`. added another `ignore` entry to filter that out since it appears to be harmless.

pllim: You could ignore the numpy warnings like this

https://github.com/astropy/astropy/blob/f96d2aee1b8c29f23f0bdd8a487580f509ca1b6f/setup.cfg#L134-L135

specreduce/tests/test_extinction.py
```diff
@@ -58,12 +61,17 @@ def test_missing_extinction_unit():
     """
     wave = np.linspace(0.3, 2.0, 50)
     extinction = 1. / wave
-    ext = AtmosphericExtinction(extinction=extinction, spectral_axis=wave * u.um)
-    assert(len(ext.extinction_mag) > 0)
-    assert(len(ext.transmission) > 0)
+    with pytest.warns(AstropyUserWarning):
+        ext = AtmosphericExtinction(extinction=extinction, spectral_axis=wave * u.um)
+        assert(len(ext.extinction_mag) > 0)
+        assert(len(ext.transmission) > 0)
```
pllim, line 67:
```suggestion
    assert len(ext.extinction_mag) > 0
    assert len(ext.transmission) > 0
```

specreduce/tests/test_extinction.py
```diff
@@ -58,12 +61,17 @@ def test_missing_extinction_unit():
...
-    ext = AtmosphericTransmission()
-    assert(len(ext.extinction_mag) > 0)
-    assert(len(ext.transmission) > 0)
+    """
+    Test creating of default atmospheric transmission model
+    """
+    with pytest.warns(RuntimeWarning):
+        ext = AtmosphericTransmission()
+        assert(len(ext.extinction_mag) > 0)
+        assert(len(ext.transmission) > 0)
```
pllim, line 77:
```suggestion
    assert len(ext.extinction_mag) > 0
    assert len(ext.transmission) > 0
```

setup.cfg
tepickering: yep, this is what i ended up doing.

specreduce/tests/test_extinction.py
tepickering: i have made the edits to do this. only the lines emitting the warnings are now within the context handlers.

specreduce/tests/test_extinction.py
tepickering: thanks for the reminder to fix the `assert` syntax. that was bugging me, but hadn't gotten around to dealing with it. fixed now.

specreduce/tests/test_extinction.py
```diff
@@ -58,12 +61,19 @@ def test_missing_extinction_unit():
...
 def test_transmission_model():
+    """
+    Test creating of default atmospheric transmission model
+    """
     ext = AtmosphericTransmission()
-    assert(len(ext.extinction_mag) > 0)
-    assert(len(ext.transmission) > 0)
+    assert len(ext.transmission) > 0
+
+    with pytest.warns(RuntimeWarning):
```
pllim, line 78: Huh, that's weird. Why would the `len()` give RuntimeWarning here?

tox.ini
```diff
@@ -47,22 +47,32 @@ deps =
     numpy120: numpy==1.20.*
     numpy121: numpy==1.21.*
 
-    astropylts: astropy==4.0.*
+    astropy50: astropy==5.0.*
+    astropylts: astropy==5.0.*
+    astropyrc: astropy==5.1rc1
```
pllim, line 52: @bsipocz has set up a way to automatically pick up astropy prerelease over at astropy/astroquery#2404 if you are interested. FYI.

specreduce/tests/test_extinction.py
tepickering: the `extinction_mag` property in that case is doing a `log10` operation that generates that warning because there are wavelengths with 0 transmission.

tox.ini
tepickering: oh cool. i'll take a look.

specreduce/tests/test_extinction.py
pllim: I wonder if this means this warning ignore needs to be moved into the property or the whole code needs some refactoring but I guess that is out of scope here. Thanks for the clarification!

pllim (approved):
LGTM!

specreduce/tests/test_extinction.py
tepickering: yeah, it's a bigger question how to best handle zero/negative values in the presence of logarithms. i think, though, one will always want a warning when an operation on an array of valid numbers generates an array with some `inf`/`nan` values. so it's a matter of whether to leave it as the original `numpy` warning or wrap it into a custom one.

tox.ini
tepickering: yeah! i like that approach and implemented it as one of the cron tests.

PR #114, opened 2022-06-29
HorneExtract: require variance array to be fully positive

### LINKED ISSUE #103: HorneExtract: variance of zeros fails
passing zeros into variance results in inf/inf which returns an extracted spectrum of 0s.  These should either be handled gracefully or raise an error.  (cc @orifox)

kecnry:
This PR requires the input `variance` to HorneExtract to be fully positive, and raises an error if it is not.  As an exception, the case of _all_ zeros, is treated as unweighted by internally defaulting to `variance = np.ones_like(variance)`.  In the future, we may want more advanced logic for handling cases with some values being zero, but for now leave it to the user to adjust the input `variance` and/or `mask`.

Closes #103

### DIFF
--- a/specreduce/tests/test_extract.py
+++ b/specreduce/tests/test_extract.py
@@ -87,12 +87,19 @@ def test_horne_variance_errors():
     # all zeros are treated as non-weighted (give non-zero fluxes)
     err = np.zeros_like(image)
     mask = np.zeros_like(image)
-    ext = extract(image.data, trace, variance=err, mask=np.zeros_like(err), unit=u.Jy)
+    ext = extract(image.data, trace, variance=err, mask=mask, unit=u.Jy)
     assert not np.all(ext == 0)
 
-    # single non-positive value raises error
+    # single zero value adjusts mask (does not raise error)
     err = np.ones_like(image)
     err[0] = 0
+    mask = np.zeros_like(image)
+    ext = extract(image.data, trace, variance=err, mask=mask, unit=u.Jy)
+    assert not np.all(ext == 0)
+
+    # single negative value raises error
+    err = np.ones_like(image)
+    err[0] = -1
     mask = np.zeros_like(image)
     with pytest.raises(ValueError, match='variance must be fully positive'):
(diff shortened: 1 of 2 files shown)


tepickering:
thanks for this. looks good to me, though i do concur with `codecov` that the new lines should be covered by tests to make sure this fix works as intended.

kecnry:
I added tests for the logic added in this PR.  Note that the existing tests did not include any coverage for `HorneExtract`, so I only added tests for the logic introduced here.  Full validation of results from `HorneExtract` should be added to the tests separately.

tepickering (approved):
looks good to me! thanks!

specreduce/tests/test_extract.py
```diff
@@ -77,3 +78,28 @@ def test_array_trace():
...
+    err[0] = 0
+    mask = np.zeros_like(image)
+    ext = extract(image.data, trace, variance=err, mask=mask, unit=u.Jy)
+    assert not np.all(ext == 0)
+
+    # single negative value raises error
+    err = np.ones_like(image)
+    err[0] = -1
+    mask = np.zeros_like(image)
+    with pytest.raises(ValueError, match='variance must be fully positive'):
```
ojustino, line 104: Hinging the test on the exact phrase used in the traceback may be dangerous if it's edited innocuously in the future. I assume this scenario would then always pass, even if the `ValueError` was triggered for the reason being tested. Should there be a comment by the phrase in `extract.py`, or is there a more canonical way to address this concern?

ojustino (approved):
This implementation also works for me. I have a comment about the test, but I consider it more something to think about than something worthy of blocking a merge.

specreduce/tests/test_extract.py
kecnry: this line _requires_ the indented block to raise a value error with that exact message or else the test will fail.  So if we change the error message at any point, we will also need to update the test (or else the test will fail).  Is that what you meant or is there another situation to worry about?

PR #115, opened 2022-07-14
Kosmos trace peak_method options

kecnry:
This pull request implements an option `peak_method` argument to KosmosTrace, which can currently be one of: 'gaussian' (default), 'centroid', or 'max' (currently unsmoothed).  See the new sandbox notebook for usage examples and comparisons.


PatrickOgle (approved):
Tried the 3 methods on a real spectrum, and they appear to work as advertised.  peak_method ='centroid' gives the best result for a source with an asymmetric spatial profile.

PatrickOgle:
Example trace with peak_method = "max":
Screen Shot 2022-07-27 at 11 17 44 AM

tepickering (changes requested):
thank you for this! very useful to have different ways to define the source center. the changes look good, but need test coverage. i'll look into why codecov isn't being run as a check on PRs...

kecnry:
@tepickering - basic test coverage added (I'm open to suggestions for something more robust, but apart from reproducing the logic to test or testing against current exact values, I don't have any great ideas).  I meant to do this before opening it as ready for review, so sorry about that!

tepickering:
no worries. what you added is fine for now. i did find and fix one issue with the config on the codecov.io end. not sure why it refused to upload the report from this PR, though. if it's related to the issue i found, it might start working when the next commits are pushed here. at any rate, it looks like once things merge everything comes out right. we'll see what happens and i'll tackle it as needed in a different PR.

tepickering (approved):

ojustino (approved):
These are good changes that all work for me. The Gaussian peak method appears to work as it did before. As @kecnry pointed out in a separate discussion, it would be great to not do all of this in one function, but a structural change is out of scope for now.

I won't let it prevent an approval, but I have a couple of suggestions for the notebook.

First, using the `aspect` argument with a value greater than 1 would help the `imshow` plot look less squished to my eye.

Second, I think a comparison plot with some or all of the traces on one figure would be a good conclusion. Here's something I worked out that strips out all of the image besides the extraction window and recalculates the normalization based on the new image so the window's peak is more highly contrasted:

```
yrange = [ext_center - ext_width,
          ext_center + ext_width]

image_window_only = np.zeros_like(image)
image_window_only[yrange[0]:yrange[1]] = image[yrange[0]:yrange[1]]

plt.figure(figsize=(10, 10))
plt.imshow(bg.sub_image(image_window_only),
           norm=simple_norm(image_window_only, 'sqrt'),
           origin="lower", aspect=20, cmap='magma')

plt.plot(auto_trace.trace, label='gaussian')
plt.plot(auto_trace_centroid.trace, label='centroid')
plt.plot(auto_trace_max.trace, label='max')
plt.plot(auto_trace_window.trace, alpha=.5, label='window')

plt.ylim(*yrange)
plt.legend()
```

If you like this, you may want to test some different colors for the traces. It's already implied  in the plot that subtracts one from the other, but it's probably good to mention again that the Gaussian and window traces are nearly indistinguishable in this joint plot at the scale chosen above.

kecnry:
Thanks for the reviews!

@ojustino - those are great suggestions - I left a note to account for them when we write up an official example notebook (for now I just wanted to show the syntax and that it produces reasonable results in the sandbox notebook).

PR #117, opened 2022-08-09
extract: more consistent API compared to other steps

kecnry:
This PR refactors the extract classes to allow initializing the dataclass and/or calling the call method.

Before:

```
ext = specreduce.extract.BoxcarExtract()
spectrum = ext(img, trace, width)
```

After:

```
ext = specreduce.extract.BoxcarExtract(img, trace)
spectrum = ext(width)
```

or

```
ext = specreduce.extract.BoxcarExtract(img, trace, width=5)
ext.width = 10
spectrum = ext.spectrum # shortcut to ext() in order to be consistent with Background
```

Note that `img` and `trace` are now required arguments when initializing the object, but any parameter can be set or changed when initializing, by setting the attribute, or when calling.

**TODO before merging**:
- [x] update examples to always pass `img`/`trace` or change to not require anything so that the API does not change
- [ ] if we do require `img`/`trace` this breaks public API so might need a version bump

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -1,16 +1,16 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 import warnings
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 import numpy as np
 
 from astropy import units as u
-from astropy.modeling import models, fitting
+from astropy.modeling import Model, models, fitting
 from astropy.nddata import NDData
 
 from specreduce.core import SpecreduceOperation
-from specreduce.tracing import FlatTrace
+from specreduce.tracing import Trace, FlatTrace
 from specutils import Spectrum1D
 
 __all__ = ['BoxcarExtract', 'HorneExtract', 'OptimalExtract']
@@ -84,8 +84,8 @@ class BoxcarExtract(SpecreduceOperation):
     Example: ::
 
         trace = FlatTrace(image, trace_pos)
-        extract = BoxcarExtract()
-        spectrum = extract(image, trace, width)
+        extract = BoxcarExtract(image, trace)
+        spectrum = extract(width=width)
 
 
     Parameters
@@ -106,11 +106,19 @@ class BoxcarExtract(SpecreduceOperation):
     spec : `~specutils.Spectrum1D`
         The extracted 1d spectrum expressed in DN and pixel units
     """
-
+    image: NDData
+    trace_object: Trace
+    width: float = 5
+    disp_axis: int = 1
+    crossdisp_axis: int = 0
     # TODO: should disp_axis and crossdisp_axis be defined in the Trace object?
 
-    def __call__(self, image, trace_object, width=5,
-                 disp_axis=1, crossdisp_axis=0):
+    @property
+    def spectrum(self):
+        return self.__call__()
+
+    def __call__(self, image=None, trace_object=None, width=None,
+                 disp_axis=None, crossdisp_axis=None):
         """
         Extract the 1D spectrum using the boxcar method.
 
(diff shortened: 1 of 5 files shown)


ojustino:
I like this idea, and the consistency argument makes sense. A side benefit here is that you can now see the full docstring on `HorneExtract` itself in an interpreter or notebook (e.g., `HorneExtract?`) without first creating a class instance.

I agree on the need for a version bump if and when this is implemented.

Instead of duplicating the text, would it be possible to programmatically grab the class docstring (sans the "[default: X]" for each argument) for each extraction class and attach it to the respective `__call__()` method?

specreduce/extract.py
```diff
@@ -106,11 +106,19 @@ class BoxcarExtract(SpecreduceOperation):
     spec : `~specutils.Spectrum1D`
         The extracted 1d spectrum expressed in DN and pixel units
     """
-
+    image: NDData
+    trace_object: Trace
+    width: float = 5
+    disp_axis: int = 1
+    crossdisp_axis: int = 0
```
ojustino, line 113: 1. Should these defaults be noted in the `BoxcarExtract` docstring? They are in `HorneExtract`.
2. The API example in the `BoxcarExtract` docstring needs an update to match the new API introduced here.

ojustino (commented):
There is another occurrence of `BoxcarExtract()` without arguments in `notebook_sandbox/jwst_boxcar/jwst_boxcar_algorithm.ipynb`, but I think that notebook was working with the pre-1.0.0 version of that class. Should it be updated or left alone as an artifact? If the latter, it probably needs a warning at the top, which could be added in another pull request.

Besides that and my comments on the `BoxcarExtract` docstring, I still think this looks good.

specreduce/extract.py
kecnry: good catch, done!

kecnry:
> There is another occurrence of BoxcarExtract() without arguments in notebook_sandbox/jwst_boxcar/jwst_boxcar_algorithm.ipynb, but I think that notebook was working with the pre-1.0.0 version of that class. Should it be updated or left alone as an artifact? If the latter, it probably needs a warning at the top, which could be added in another pull request.

That's a good question.  I agree that was somewhat the "concept" notebook and has not been kept up to date with the actual specreduce API.  I like the idea of a warning at the top, and agree that might be best done as a separate PR.

> Besides that and my comments on the BoxcarExtract docstring, I still think this looks good.

I do like the docstring idea *if* it can be done cleanly and isn't too confusing.  If you have something in mind that works, feel free to push directly to this PR.  Otherwise, I think we can implement it down the road at any point (I'd just like to keep this moving if possible, since the decision on the API here is blocking work in jdaviz).

ojustino (approved):
I am fine with merging, then. I'll create an issue for adding warnings to the top of the sandbox notebooks and submit another pull request if my idea for the `__call__()` docstrings works.

specreduce/extract.py
```diff
@@ -134,6 +142,12 @@ def __call__(self, image, trace_object, width=5,
             The extracted 1d spectrum with flux expressed in the same
             units as the input image, or u.DN, and pixel units
         """
+        image = image if image is not None else self.image
```
tepickering, line 145: not a showstopper or blocker by any means, but there should be a more elegant way to handle this pattern. i'll give it a think.

tepickering (approved):
these changes look good and do improve the usability of the API. i agree with codecov's nitpick about the new line not being covered by a test. once that's sorted, it's gtg to merge.

specreduce/extract.py
kecnry: that would definitely be nice. I'm not sure if dataclass has anything built-in. If not, we could implement a simple decorator to parse input arguments and fallback on the set defaults?

tepickering: actually, shouldn't it work to just set the default values to the `self.*` ones rather than `None`? that's effectively what you're doing, anyway.

tepickering (approved):

specreduce/extract.py
kecnry: I think that would require a decorator, I don't think `def __call__(self, image=self.image, trace_object=self.trace_object, ...)` is valid syntax, is it?  Or is there some other way to handle defaults?

pllim:
I think this broke Jdaviz...

PR #118, opened 2022-08-15
Correct default mask for array images in HorneExtract

ojustino:
When `HorneExtract`/`OptimalExtract` is given a `numpy` array as an image and no mask, it tries to create its own mask using `np.ma.masked_invalid(image)`. They are currently written as if that object itself is the mask, when it's actually in that object's `mask` attribute. This pull request makes the switch.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,8 +10,13 @@ API Changes
 ^^^^^^^^^^^
 
 - ``BoxcarExtract`` and ``HorneExtract`` now accept parameters (and require the image and trace)
-  at initialization, and allow overriding any input parameters when calling. [#117]
+  at initialization, and allow overriding any input parameters when calling [#117]
 
+Bug Fixes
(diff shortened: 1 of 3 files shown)


ojustino:
Good call, @kecnry; I went ahead and created a new section in `CHANGES.rst`.

tepickering:
looks good and another nice catch. codecov says and i agree it'd be worth adding a test case to cover this.

ojustino:
The coverage needed here likely clashes with #117 since `image` is one of the parameters whose location is changed there. That one is close to complete, though we may need to discuss how to handle a merge given that it breaks the API.

kecnry:
@ojustino - if you add a test here that covers the case that used to fail, I can update it if needed when rebasing (since I'm guessing this will get merged before that one).

ojustino:
OK, the tests are in.

PR #119, opened 2022-08-15
Correct license path in README

ojustino:
I noticed the path to the license in the README went Commonwealth-style for a moment.


tepickering:
haha. habits formed from living in the Commonwealth sometimes leak out. nice catch and thank you.

kecnry (approved):

PR #110, opened 2022-06-02
Add citation and Zenodo badge

tepickering:
A Zenodo DOI has been created for the `1.0.0` release and a badge pointing to that added to the `README` file. A `CITATION.cff` file has been added here that points to that DOI. I have thus far only added myself as an author in that file. I encourage other contributors to add their information to it either by editing the file directly in this PR or adding it later in a separate PR. 

Closes #108.

PR #120, opened 2022-08-15
Update citation file

ojustino:
I missed #110 and am rectifying it here. I'm assuming @tepickering and @eteq are respectively first and last on purpose, with the middle going in alphabetical order. Do we have such a system for this file?

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -20,6 +20,10 @@ authors:
     family-names: Conroy
     affiliation: STScI
     orcid: 'https://orcid.org/0000-0002-5442-8550'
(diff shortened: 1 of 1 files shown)


tepickering:
thanks for this! we never really discussed a system. i'm fine with leaving it as you suggest or just make it purely alphabetical.

ojustino:
There are probably more people to add, so it might be more efficient to decide once the list is more complete. Another approach to consider at that juncture is to make the first author something like "Specreduce Developers" and then list everyone else alphabetically.

kecnry (approved):

PR #124, opened 2022-08-18
Preparation for release v1.1.0

ojustino:
I applied the version's tag here, though doing it in #123 would have saved a step.


kecnry (approved):

ojustino:
For posterity's sake, adding the tag in a pull request is not the way to go.

Issue #101, opened 2022-04-07
compatibility issues between KosmosTrace and Background


kecnry:
modified script to reproduce from the `boxcar_extraction.ipynb` notebook:

```
from pathlib import Path
from zipfile import ZipFile

from astropy.io import fits
from astropy.table import Table
from astropy.utils.data import download_file

from jwst import datamodels

from specreduce.extract import BoxcarExtract
from specreduce.tracing import FlatTrace, KosmosTrace
from specreduce.background import Background

import tempfile

zipped_datapath = Path(download_file('https://stsci.box.com/shared/static/qdj79y7rv99wuui2hot0l3kg5ohq0ah9.zip', cache=True))
data_dir = Path(tempfile.gettempdir())
with ZipFile(zipped_datapath, 'r') as sample_data_zip:
    sample_data_zip.extractall(data_dir)

s2dfile = str(data_dir / "nirspec_fssim_d1_s2d.fits")
s2d = datamodels.open(s2dfile)
image = s2d.slits[0].data

# this works:
trace = FlatTrace(image, 27)
bg = Background.two_sided(image, trace, separation=5, width=2)

# this raises an error that the background regions overlap (even though they shouldn't)
trace = KosmosTrace(image, 27)
bg = Background.two_sided(image, trace, separation=5, width=2)

# this raises an integer error because the trace goes out of bounds of the image:
trace = KosmosTrace(image, 27)
bg = Background.two_sided(image, trace, separation=6, width=2)
```

The integer error is a pretty easy fix (we can avoid the mask by acting on `trace.data` instead of `trace`), but we have to decide whether the background trace going out of bounds should raise an error, raise an error only if the entire boxcar falls outside the image, or continues without an error (and just includes pixels in the boxcar if within range and zeros or nans otherwise).

ojustino:
I also saw that the overlap error affected more than just backgrounds created with `KosmosTrace`, so I don't find this unexpected. I don't get an error when I run this line on that PR's branch, though I do have to enclose the `FlatTrace` in a list.

ojustino:
We have resolved this issue by merging #125 and #127.

PR #125, opened 2022-08-19
Found cause of background "overlap" error

ojustino:
Addresses the overlapping background regions error described in part of #101.

I think the real issue lies with `_get_boxcar_weights()`, not `Background` or its interaction with `KosmosTrace`. In short, that function gives some pixels weights above 1, and `Background` assumes that any pixel with a value above 1 got that way due to overlapping regions. That's not the case here.

For example, take the image from the code block in #101. The aperture center of the image's first column as found by `KosmosTrace` is 26.79. The code snippet referencing the overlap error sets the background's separation to 5, so the first `center` given to `_get_boxcar_weights()` is 21.79. The half-width of the aperture defined in the snippet is 1 pixel, and the image is 34 pixels across in its cross-dispersion direction. On `main`, this combination of inputs gives a peak weight above 1:

```
# _get_boxcar_weights(center=21.79, hwidth=1, npix=34) (on main)
array([0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ,
       0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.71,
       **1.29**, 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ,
       0.  ])
```

Using the modified function in this branch, the peak weight is 1:

```
# _get_boxcar_weights(center=21.79, hwidth=1, npix=34) (in PR)
array([0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ,
       0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.21, **1. ** ,
       0.79, 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ,
       0.  ])
```
<br />

For a simpler example, define the aperture `center` as 27.5 and `hwidth` as 0.5. I expect all the weight to fall onto one pixel, the 27th (zero-indexed). That's what I see with my modifications:

```
# _get_boxcar_weights(center=27.5, hwidth=0.5, npix=34) (in PR)
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., **1. **, 0., 0., 0., 0., 0., 0.])
```

On `main`, the weights are instead divided evenly onto the 27th and 28th pixels:
```
# _get_boxcar_weights(center=27.5, hwidth=0.5, npix=34) (on main)
array([0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. ,
       0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. , 0. ,
       0. , 0.5, 0.5, 0. , 0. , 0. , 0. , 0. ])
```
<br />

Does this make sense to others? If so, are there any ramifications for `BoxcarExtract` since we thought it was working properly?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,9 @@
 Bug Fixes
 ^^^^^^^^^
 
-- Improved errors/warnings when background region extends beyond bounds of image. [#127]
-
+- Improved errors/warnings when background region extends beyond bounds of image [#127]
+- Fixed boxcar weighting bug that often resulted in peak pixels having weight
+  above 1 and erroneously triggered overlapping background errors [#125]
 
 1.1.0
 -----
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -59,14 +59,10 @@ def _get_boxcar_weights(center, hwidth, npix):
         w0 = hwidth - (center - fullpixels[0])
         if w0 >= 0:
             weights[fullpixels[0] - 1] = w0
-        else:  # does this scenario happen anymore given the adjustment to w0?
-            weights[fullpixels[0]] = 1 + w0
     if fullpixels[1] < npix:
         w1 = hwidth - (fullpixels[1] - center)
         if w1 >= 0:
             weights[fullpixels[1]] = w1
-        else:  # does this scenario happen anymore given the adjustment to w1?
-            weights[fullpixels[1]] = 1 + w1
 
     return weights
 
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -45,6 +45,10 @@ def _get_boxcar_weights(center, hwidth, npix):
     """
     weights = np.zeros(npix)
 
+    # shift center from integer to pixel space, where pixel N is [N-0.5, N+0.5),
+    # not [N, N+1). a pixel's integer index corresponds to its middle, not edge
+    center += 0.5
+
     # pixels given full weight because they sit entirely within the aperture
     fullpixels = [max(0, int(np.ceil(center - hwidth))),
                   min(int(np.floor(center + hwidth)), npix)]
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -56,13 +56,9 @@ def _get_boxcar_weights(center, hwidth, npix):
 
     # pixels at the edges of the boxcar with partial weight, if any
     if fullpixels[0] > 0:
-        w0 = hwidth - (center - fullpixels[0])
-        if w0 >= 0:
-            weights[fullpixels[0] - 1] = w0
+        weights[fullpixels[0] - 1] = hwidth - (center - fullpixels[0])
     if fullpixels[1] < npix:
-        w1 = hwidth - (fullpixels[1] - center)
-        if w1 >= 0:
-            weights[fullpixels[1]] = w1
+        weights[fullpixels[1]] = hwidth - (fullpixels[1] - center)
 
     return weights
 
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -55,13 +55,13 @@ def _get_boxcar_weights(center, hwidth, npix):
         w0 = hwidth - (center - fullpixels[0])
         if w0 >= 0:
             weights[fullpixels[0] - 1] = w0
-        else: # does this scenario happen given adjustment made to w0?
+        else:  # does this scenario happen anymore given the adjustment to w0?
             weights[fullpixels[0]] = 1 + w0
     if fullpixels[1] < npix:
         w1 = hwidth - (fullpixels[1] - center)
         if w1 >= 0:
             weights[fullpixels[1]] = w1
-        else: # does this scenario happen given adjustment made to w1?
+        else:  # does this scenario happen anymore given the adjustment to w1?
             weights[fullpixels[1]] = 1 + w1
 
     return weights
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -19,24 +19,50 @@ __all__ = ['BoxcarExtract', 'HorneExtract', 'OptimalExtract']
 def _get_boxcar_weights(center, hwidth, npix):
     """
     Compute weights given an aperture center, half width,
-    and number of pixels
+    and number of pixels.
+
+    Based on `get_boxcar_weights()` from a JDAT Notebook by Karl Gordon:
+    https://github.com/spacetelescope/jdat_notebooks/blob/main/notebooks/MIRI_LRS_spectral_extraction/miri_lrs_spectral_extraction.ipynb
+
+    Parameters
+    ----------
+    center : float, required
+        The index of the aperture's center pixel on the larger image's
+        cross-dispersion axis.
+
+    hwidth : float, required
+        Half of the aperture's width in the cross-dispersion direction.
+
+    npix : float, required
+        The number of pixels in the larger image's cross-dispersion
+        axis.
+
+    Returns
+    -------
+    weights : `~numpy.ndarray`
+        A 2D image with weights assigned to pixels that fall within the
+        defined aperture.
     """
-    weights = np.zeros((npix))
+    weights = np.zeros(npix)
 
-    # pixels with full weight
-    fullpixels = [max(0, int(center - hwidth + 1)),
-                  min(int(center + hwidth), npix)]
-    weights[fullpixels[0]:fullpixels[1]] = 1.0
+    # pixels given full weight because they sit entirely within the aperture
+    fullpixels = [max(0, int(np.ceil(center - hwidth))),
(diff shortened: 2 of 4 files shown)


kecnry:
I'll need to carefully think through the exact logic, but in general I think this makes sense.  Weights should never be above 1 for a single boxcar, and so the check for overlapping regions _should_ be able to remain once this is fixed.  This would also probably mean that the previous implementation was giving wrong results, depending on where exactly mid-pixel the boxcar fell... so we'll need to carefully update and validate CI.  A notebook showing plots of these weights along with lines of where the boxcar falls on the pixels for different rounding situations would be _super_ helpful, both for us to write tests and to illustrate the behavior to users.

ojustino:
This does affect `BoxcarExtract` (the source of the failed tests), but I have a better idea of what's going on. [My original post](https://github.com/astropy/specreduce/pull/125#issue-1344472954) demonstrates that `_get_boxcar_weights()` does not work as expected in `main` when `center` isn't a whole number.

After a suggestion from @kecnry, I also checked cases where `center` **is** a whole number. The difference between my approach and `main` in those scenarios is that mine assumes the center is at the "bottom" edge of the pixel while `main` assumes it's in the middle. With `center=X, hwidth=0.5`, my approach puts a weight of 0.5 each on pixels X-1 and X while `main` puts full weight on pixel X. `main`'s approach in this scenario actually seems more intuitive to me, especially in a context like Specviz2D's extraction plugin.

`main`'s approach can work all the time if we necessitate that trace centers always be whole numbers (with the understanding that centering on pixel X really means centering on the middle of it, or position `X + 0.5`). If that's not OK, do we want my approach to replicate `main`'s behavior when `center` is a whole number? If so, we could check on whether `center` is an int or a float, but then we'd get different weights with `center=10` and `center=10.0`, for example.

I don't mind creating a notebook, but I think it should happen once we're more settled on the weighting algorithm.

kecnry:
I do understand how the separate cases you propose feel most intuitive that way, but changing the behavior based on the data type passed concerns me.  After thinking about this and discussing it offline, I think we should always define the `N.0` as being the center of the pixel (regardless whether its a float or integer), meaning that a pixel spans from `N-1+0.5` to `N+0.5`.  I've found the most intuitive way to think about this practically is the different between rounding and flooring.  Looking at a few examples:

`center=5.0, half_width=0.5`: 
* pixel `floor(N)=5` gets weight 1

`center=5.0, half_width=1.0`:
* `floor(N)+1=6` gets weight 0.5
* `floor(N)=5` gets weight 1
* `floor(N)-1=4` gets weight 0.5

`center 5.7, half_width=1.0`:
* `floor(N)+2=7` gets weight 0.2
* `floor(N)+1=6` gets weight 1.0 (center actually falls on this pixel since `round(5.7) = floor(5.7)+1`)
* `floor(N)=5` gets weight 0.8

this last one originally feels unintuitive since you're passing a center of 5.7 with a half_width of 1, which without thinking about the definition feels like the entirety of pixel 5 should be included.  But under this definition, the pixel is actually defined as center=4.5-5.5, so the half_width falls short by 0.2.

I think this is the only way that we can keep the intuitive nature of integer pixels while also supporting partial pixels (which is needed for KosmosTrace, but will also be needed when we add support for partial-pixel widths).

PatrickOgle:
Not sure I understand the issue completely, but we do need to allow fractional extraction centers and trace locations.

tepickering:
i will note that `photutils` follows the convention where the pixel indices denote the pixel centers. thus, the pixel at `[0,0]` spans `[-0.5:0.5,-0.5:0.5]`: https://photutils.readthedocs.io/en/stable/pixel_conventions.html. i would suggest we follow the same convention for consistency. 

we will likely also want to adopt the techniques `photutils` uses to more properly account for partial pixel coverage. the `exact` mode would be hella complex for the arbitrary aperture shapes we need to handle, but the `subpixel` mode would be doable.

ojustino:
OK, then it seems everyone (including @camipacifici and @patrickogle in a separate conversation) has converged in the same direction that says the center of pixel N is N and not N + 0.5. I will update this branch to that effect.

ojustino:
My latest commit returns the definition of the center of a pixel to what it was before – N is the center of pixel N, which spans [N-0.5, N+0.5). It still also fixes the over-weighting bug in all the cases I've tested.

I changed the expected value of a case in `test_boxcar_extraction()` (position 14.3, width 4.7) that also falls victim to the over-weighting bug in `main`. I think this because the sum of the weights on one of the columns in the dispersion direction is higher than the number of cross-dispersion pixels.

_[In other words, if each (disp, x-disp) combination can have a maximum weight of 1 and there are NN x-disp pixels, the sum of weights in a disp column shouldn't be more than NN, but it is in one column of the weights image for that position and width.]_

ojustino:
The lines that drag down codecov may not even be necessary anymore, but I would like others to at least think through whether I'm right about that before I remove them.

kecnry:
`w0 = hwidth - (center - fullpixels[0])`: `w0` can only be negative if `fullpixels[0] < center - hwidth` which is impossible given `fullpixels[0] = max(0, ceil(center-hwidth))`

similarly, `w1 = hwidth - (fullpixels[1] - center)` can only be negative if `fullpixels[1] > center + hwidth` which is impossible given `fullpixels[1] = min(int(np.floor(center + hwidth)), npix)`

So I think its safe to say that the else-statements can be safely removed and we don't need to worry about the missing coverage there since they're now impossible to enter.

kecnry (approved):
As long as we open a follow-up issue/ticket for the small width bug (which is out-of-scope of the original bug, even though its likely this same code logic that needs fixing), this looks good to me!  I tried with a bunch of test scenarios and could no longer trigger the background overlap error when it shouldn't, and still can when it should!

specreduce/extract.py
```diff
@@ -19,24 +19,50 @@
...
+    # pixels at the edges of the boxcar with partial weight, if any
     if fullpixels[0] > 0:
-        w = hwidth - (center - fullpixels[0] + 0.5)
-        if w >= 0:
-            weights[fullpixels[0] - 1] = w
-        else:
-            weights[fullpixels[0]] = 1. + w
+        w0 = hwidth - (center - fullpixels[0])
+        if w0 >= 0:
+            weights[fullpixels[0] - 1] = w0
```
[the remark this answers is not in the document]
ojustino: I accepted both suggestions but did so manually since I personally wanted to test the process for adding a co-author from the command line.

ojustino:
@kecnry is correct that neither `main` nor my branch are always correct when `width` is less than a pixel. Since this branch does still solve #101 and that bug is holding some other work back, I propose we merge for now as long as everyone is happy with its current state. I will create a follow-up issue about a fix for sub-pixel width situations either way.

CHANGES.rst
```diff
@@ -1,6 +1,15 @@
-1.1.0 (unreleased)
+1.2.0 (unreleased)
 ------------------
 
+Bug Fixes
+^^^^^^^^^
+
```
tepickering, line 6: maybe add mention of fixing pixel coordinate convention

tepickering (changes requested):
looks good to me. before merging i would add a bit to the change log to mention changing pixel coordinate convention to reference the center vs a corner. codecov is fussing about a slight drop in coverage as well so want to address that as well.

CHANGES.rst
ojustino: I was saying [here](https://github.com/astropy/specreduce/pull/125#issuecomment-1227499628) that the convention hasn't changed, it just wasn't made explicit in `_get_boxcar_weights()` until we got into the weeds of why the overweighting issue occurred. As such, I think mentioning it in the changelog would cause confusion.

tepickering: ah, sorry i missed/forgot that. you are right. leave as-is here.

ojustino:
@tepickering, it's not showing up after I resolved the changelog's merge conflict, but codecov passed on the penultimate commit. I'm not sure why it's failing now – the if/else block I edited was already there.

tepickering:
from the coverage report it looks like the miss that's tripping the check is the `FlatTrace` position check, line 179.

ojustino:
I still don't see how these changes affect coverage for the lines indicated in the report, but I added a couple of tests for `KosmosTrace` so we can merge this soon.

tepickering:
well, coverage has notched up a bit so the changes worked.

tepickering (approved):

PR #127, opened 2022-08-25
logic to detect background regions extending beyond image limits

kecnry:
This PR adds logic to check if a background region extends beyond the image limits.  In the case where the center of the window (of any of the background traces) falls off the image at any dispersion "column", an error is raised.  If any part of the window falls off the image, a warning is raised.

This (in addition to #125) should address #101.

Following the case in #101:

```
from pathlib import Path
from zipfile import ZipFile

from astropy.io import fits
from astropy.table import Table
from astropy.utils.data import download_file

from jwst import datamodels

from specreduce.extract import BoxcarExtract
from specreduce.tracing import FlatTrace, KosmosTrace
from specreduce.background import Background

import tempfile

zipped_datapath = Path(download_file('https://stsci.box.com/shared/static/qdj79y7rv99wuui2hot0l3kg5ohq0ah9.zip', cache=True))
data_dir = Path(tempfile.gettempdir())
with ZipFile(zipped_datapath, 'r') as sample_data_zip:
    sample_data_zip.extractall(data_dir)

s2dfile = str(data_dir / "nirspec_fssim_d1_s2d.fits")
s2d = datamodels.open(s2dfile)
image = s2d.slits[0].data
```

the following used to raise a cryptic integer error, but now raises a helpful error message ("center of background window goes beyond image boundaries"):
```
bg = Background.two_sided(image, trace, separation=7, width=2)
```

and the following now raises a warning ("background window extends beyond image boundaries"):
```
bg = Background.two_sided(image, trace, separation=6, width=2)
```

this last example currently raises an error that the regions overlap, but testing this branch locally on top of #125 should show that this case succeeds in creating the background after raising the warning.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
-1.1.0 (unreleased)
+1.2.0 (unreleased)
 ------------------
+
+Bug Fixes
+^^^^^^^^^
+
+- Improved errors/warnings when background region extends beyond bounds of image. [#127]
+
+
+1.1.0
+-----
 
 New Features
 ^^^^^^^^^^^^
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import warnings
 from dataclasses import dataclass, field
 
 import numpy as np
@@ -87,6 +88,13 @@ class Background:
         bkg_wimage = np.zeros_like(self.image, dtype=np.float64)
         for trace in self.traces:
             trace = _to_trace(trace)
+            if (np.any(trace.trace.data >= self.image.shape[self.crossdisp_axis]) or
+                    np.any(trace.trace.data < 0)):
+                raise ValueError("center of background window goes beyond image boundaries")
+            elif (np.any(trace.trace.data + self.width/2. >= self.image.shape[self.crossdisp_axis])
+                  or np.any(trace.trace.data - self.width/2. < 0)):
+                warnings.warn("background window extends beyond image boundaries")
+            # pass trace.trace.data to ignore any mask on the trace
             bkg_wimage += _ap_weight_image(trace,
                                            self.width,
                                            self.disp_axis,
--- a/specreduce/tracing.py
+++ b/specreduce/tracing.py
@@ -51,7 +51,7 @@ class Trace:
             Shift to be applied to the trace
         """
         # act on self.trace.data to ignore the mask and then re-mask when calling _bound_trace
-        self.trace = self.trace.data + delta
+        self.trace = np.asarray(self.trace.data) + delta
         self._bound_trace()
 
     def _bound_trace(self):
--- a/specreduce/tracing.py
+++ b/specreduce/tracing.py
@@ -50,7 +50,8 @@ class Trace:
         delta : float
             Shift to be applied to the trace
         """
-        self.trace += delta
+        # act on self.trace.data to ignore the mask and then re-mask when calling _bound_trace
+        self.trace = self.trace.data + delta
         self._bound_trace()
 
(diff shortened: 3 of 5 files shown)


specreduce/tracing.py
```diff
@@ -50,7 +50,8 @@ def shift(self, delta):
         delta : float
             Shift to be applied to the trace
         """
-        self.trace += delta
+        # act on self.trace.data to ignore the mask and then re-mask when calling _bound_trace
+        self.trace = self.trace.data + delta
```
kecnry, line 54: this seems to work locally for me, but on CI is giving `TypeError: unsupported operand type(s) for +: 'memoryview' and 'int'`.

kecnry: ("fixed" by casting with `np.asarray`)

tepickering (approved):
nice catch and looks good to me!

CHANGES.rst
```diff
@@ -1,6 +1,15 @@
-1.1.0 (unreleased)
+1.2.0 (unreleased)
 ------------------
 
+Bug Fixes
+^^^^^^^^^
+
+- Improved errors/warnings when background region extends beyond bounds of image.
```
ojustino, line 7:
```suggestion
- Improved errors/warnings when background region extends beyond bounds of
  image [#127]
```

specreduce/background.py
```diff
@@ -87,6 +88,13 @@ def _to_trace(trace):
         bkg_wimage = np.zeros_like(self.image, dtype=np.float64)
         for trace in self.traces:
             trace = _to_trace(trace)
+            if (np.any(trace.trace.data >= self.image.shape[self.crossdisp_axis]) or
+                    np.any(trace.trace.data < 0)):
+                raise ValueError("center of background window goes beyond image boundaries")
```
ojustino, line 93: Should an error only be raised when `all` (instead of `any`) of the center of the background window is out of bounds? I assume `any` triggering an error instead of a warning could be annoying in a situation where only a couple of values are off-frame due to a wiggly input trace.

If we switched to `all`, we'd have to confirm that any trace values passed to `_ap_weight_image()` here are less than `self.image.shape[self.crossdisp_axis] + hwidth` in order to avoid an `IndexError` from `_get_boxcar_weights()`.

ojustino (commented):
The test case also works for me when I combine with #125. I do have a question on the proper situation for warnings versus an error when going off-frame.

specreduce/background.py
kecnry: That is an option.  The current implementation raises a warning if any of any window goes out of bounds, but an error if the _center_ of any window goes out of bounds (meaning that at least half of at least one of the windows extends out-of-bounds).  My concerns about extending this logic so the error is only raised if all of the window(s) is out-of-bounds is that the same warning will then be raised for 1-pixel out-of-bounds as only 1-pixel remaining.  And if we want to change the error to if no pixels at all are included from any of the windows, then the logic does become significantly more complex.

ojustino: If I'm understanding correctly that this branch's approach is actually more forgiving than `main`'s, that's good to know. I am interested in what others think about this topic but don't think it should delay a merge much longer if we don't hear back.

kecnry:
Loosening up the restriction to allow some (but not all) background windows to go out-of-bounds is eventually desired as well, but results in an index error in `_get_boxcar_weights`.  The intention of this PR is to provide user-friendly error messages instead of the cryptic error messages that were raised previously.  I opened #131 to make this more flexible, at which point we can loosen the checks implemented here.

ojustino (approved):
Thank you for following up on the window restriction question from my review, @kecnry. I'll leave that conversation unresolved since it might be helpful to see the details when we come back to it.

For now, as we've resolved to address that question in the future, let's merge this pull request.

PR #128, opened 2022-08-29
Introducing polynomial tracing

### LINKED ISSUE #87: KosmosTrace polynomial option
Currently, KosmosTrace returns a spline fit to the trace.  In the case of a noisy trace, a polynomial fit would be less affected by noise.

### LINKED ISSUE #121: Add warning to sandbox notebooks
Although all of our notebooks live in a "sandbox" directory, some are more suitable as up-to-date examples of `specreduce` functionality than others. Users may need clearer guidance to discern which we most recommend.

As a start, I'd say `horne_extract/optimal_extract_VLT.ipynb` and `jwst_boxcar/jwst_boxcar_algorithm.ipynb` deserve this warning because they were made before we finalized the current state of our extraction classes. I've seen an example of someone using outdated code from one of these elsewhere.

ojustino:
I went with Astropy's `modeling.models.Polynomial1D` instead of numpy's `polyfit` to keep that calculation "in-house."

Some comments and questions:
- It wasn't a prerequisite, but I tried adding the capability to bin in the x/dispersion/wavelength direction so it's possible to compare based on the combined signal across multiple dispersion pixels when calculating peak cross-dispersion indices. Is this useful? Is the implementation sensible? I'm curious to see what others think.
- If we keep that functionality, is it confusing to take `bin_size` (as in number of dispersion pixels) as an argument of `PolyTrace` when `KosmosTrace` takes the number of bins instead?
- I'm still figuring out how best to handle images with masked values.

Would fix #87.
EDIT: A subsequent commit resolves #121.


kecnry:
How much of the implementation here is different than `KosmosTrace`?  Should we just refactor to create a generic `AutoTrace` that takes any generic astropy model as one of the parameters (and then we could always have classmethods or subclasses which hardcode specific options for convenience), or is there a sufficient need for separate logic that this isn't feasible?

kecnry:
I did test swapping out `KosmosTrace` with `PolyTrace` in the sandbox notebook and it seems to run as advertised (we'll definitely need some cases covered in tests/notebooks though and I do think if we don't do a refactor, we should at least make the API as consistent as possible between this and `KosmosTrace`).

PatrickOgle:
I would like, as much as possible, for PolyTrace to do everything the same and have the inputs and outputs be the same as for KosmosTrace, with the only difference being that a polynomial is fit rather than a spline.

tepickering:
couple quick comments:

- be mindful of the coverage reports. there's are quite a few additions here that are not covered by tests.
- regular polynomials can be very badly behaved near domain edges, especially at higher orders. legendre or chebyshev polynomials are much more appropriate for the kind of interpolation task we're doing here. fortunately, the `astropy` API is quite similar for each 1D polynomial case. for reference, IRAF and PypeIt default to legendre polynomials for aperture tracing while offering chebyshev and normal polynomials (as well as splines).

PatrickOgle:
Suggest leaving Chebyshev or Legendre polynomial options for a future PR.  This polynomial trace capability is urgent so I would like to see it merged soon.

ojustino:
The most recent two commits are different approaches to polynomial tracing based on the feedback from previous comments.

The first approach keeps `PolyTrace` as its own class. Like `KosmosTrace`, it now takes `bins` as an argument instead of `bin_size`. This version makes better use of `numpy` broadcasting in calculating bins' peak pixels. I changed its ending model to `Legendre1D`. As long as the only argument used in the model is `degree`, it's trivial to switch between it, `Polynomial1D`, `Chebyshev1D` from a coding perspective. 

In the second approach, I renamed `KosmosTrace` as `AutoTrace` and added `"polynomial"` as an option for its `peak_method` argument. Because each `peak_method` has its own quirks, the approach has to resort to looping to calculate bins' peak pixels. I used `Legendre1D` to fit the peak pixels for the polynomial case; the other `peak_method` options still use a spline, as before.

The binning procedure is different in each approach. The first has the flexibility to partially weight pixels when `bins` is not a factor of the number of cross-dispersion pixels. (For 4 bins and 10 pixels, bin 0 has weights `[1, 1, 0.5]` on pixels 0-2 and bin 1 has weights `[0.5, 1, 1]` on pixels 2-4.) The second instead rounds the bin edges to integers before before proceeding to peak finding. (For 4 bins and 10 pixels, bin 0 is pixels 0-1 and bin 1 is pixels 2-4.) The latter is how `KosmosTrace` already worked.

I am aware the tests will need an update but wanted feedback on which approach this pull request should take before moving further along. (The second especially would require changes across the repository.) Which seems like the better path? Do you suggest any changes to its implementation?

kecnry:
I personally like the idea of a general `AutoTrace`, but isn't the goal of this PR to be able to swap out the spline interpolation (rather than the gaussian peak finding) with a polynomial?  So maybe a `trace_function` or `trace_model` argument that defaults to spline but allows 'polynomial', or perhaps eventually a general astropy model object, as input?

ojustino:
I implemented @kecnry's suggestion of allowing the user to pass one of a few models as an argument in both approaches. In the `AutoTrace` approach, I think this eliminates the need for a `"polynomial"` option for `peak_method` since pairing `peak_method="max
"` and `trace_method` as any of the polynomial models serves that purpose.

I do recommend that anyone skipping to the end of the thread still read [my summary of those commits](https://github.com/astropy/specreduce/pull/128#issuecomment-1263644023).

specreduce/tracing.py
```diff
@@ -177,24 +175,31 @@ class KosmosTrace(Trace):
...
+        adopts the central value as the peak. (Based on the "kosmos"
+        algorithm from James Davenport's same-named repository.)
+        ``centroid``: Takes the centroid of the window within in bin.
+        ``max``: Saves the position with the maximum flux in each bin.
+        [default: ``gaussian``]
+
+    Possible Improvements
+    ---------------------
+    1) return info about trace width (?)
+    2) add re-fit trace functionality (or break off into other method)
```
kecnry, line 197: should these be removed from the docstring and into issues/tickets?  Or are they intentionally here to be more visible to the user?

specreduce/tracing.py
```diff
@@ -177,24 +175,31 @@ class KosmosTrace(Trace):
         guess position. Useful for tracing faint sources if multiple
         traces are present, but potentially bad if the trace is
         substantially bent or warped. [default: None]
+    trace_model : `~astropy.modeling.Model` or, optional
+        The 1-D polynomial model used to fit the trace to the bins' peak
+        pixels. Valid options are ``Spline1D``, ``Legendre1D``,
+        ``Chebyshev1D``, and ``Polynomial1D``. ``Spline1D`` is used with
```
kecnry, line 181: this makes it sound like the strings are accepted.  Can we use intersphinx links to link to the astropy classes themselves?

But maybe it would also be useful to support passing the string names in which case the default options for each model would be used?  Or we could leave that up to either classmethods or subclasses:

```
AutoTrace.polynomial(image, order=2, ...)
```

or

```
PolyTrace(image, order=2, ...) # which is just a subclass of AutoTrace except that it accepts input arguments specfic to that model
```

specreduce/tracing.py
```diff
@@ -177,24 +175,31 @@ class KosmosTrace(Trace):
...
+        ``max``: Saves the position with the maximum flux in each bin.
+        [default: ``gaussian``]
+
+    Possible Improvements
+    ---------------------
+    1) return info about trace width (?)
+    2) add re-fit trace functionality (or break off into other method)
     """
-    bins: int = 20
+    bins: int = 4
```
kecnry, line 199: might want to explicitly mention this default change in the change log?  We're not too strict about breaking changes yet, so I don't think we need to deprecate first, but we should at least try to make the change as visible as possible.

specreduce/tracing.py
```diff
@@ -221,17 +226,23 @@ def __post_init__(self):
...
-
-        cols = img.shape[self._disp_axis]
-        if self.bins >= cols:
-            raise ValueError(f"bins must be < {cols}, the length of the "
-                             "image's spatial direction")
+        valid_models = ["Spline1D", "Legendre1D", "Chebyshev1D", "Polynomial1D"]
+        trace_model_name = type(self.trace_model).name
+        if trace_model_name not in valid_models:
+            raise ValueError("trace_model must be one of "
+                             f"{', '.join(valid_models)}.")
```
kecnry, line 245: is there a benefit to this over isinstance checks?  In theory, this would let some non-astropy model object with the same class name through the check

kecnry (commented):
I really like the flexibility of the `AutoTrace` implementation and would advocate for removing the separate `PolyTrace` implementation unless there is a good reason otherwise.  As I put in one of the comments, I think we could also have classmethods or subclasses to add more convenience for providing the options to the supported trace models (but that could always be a follow-up effort).

Once we commit to one of the two implementations, this will also need test coverage, updates to the concept notebooks, and (minor) updates to #142.

specreduce/tracing.py
kecnry: I can't imagine real data would likely have less than 20 dispersion pixels, but I guess our other option would be to have it based on the size of the input image (something like `max(20, npixels/4)`)

specreduce/tracing.py
ojustino: Although the non-spline options are `models.PolynomialModel` instances, so are many other Astropy models. Their most common ancestor with `Spline1D` is `astropy.modeling.core.FittableModel`, so maybe I could check for that and then check that the name of the type comes from the `valid_models` list I've already defined.

kecnry: since you need a list somewhere anyways (unless you want to open it to any generic model), I think its just as short to do:

```
from astropy.modeling import models

valid_models = (models.Spline1D, models.Legendre1D, models.Chebyshev1D, models.Polynomial1D)
if not isinstance(trace_model, valid_models):
    raise ValueError(...)
```

specreduce/tracing.py
ojustino: These came with the original code from `kosmos`. I removed a few as we added them in, but these are probably better documented as issues than in the docstring.

ojustino:
Suggestions by @PatrickOgle on behalf of our team's product owners:

<blockquote>

- There should be one user-facing tracing class named `Trace` to prevent the need to  remember variant names like `KosmosTrace`, `PolyTrace`, etc.
- The class should include capabilities for polynomial tracing. Even `FlatTrace` could be absorbed into the class since it's technically a degree 0 polynomial fit. [note from @ojustino: could `ArrayTrace` be absorbed?]
- Since our base tracing class is already named `Trace`, it should be renamed as `BaseTrace` instead.
- Given the chance to design tracing from scratch, they would not include spline fitting and the ability to bin. Since they already exist here, the new solution can include them, but they should be de-emphasized in documentation and examples.
</blockquote>

kecnry:
I still need to give this more thought, but wanted to get my initial ideas down somewhere in the meantime:

> Even `FlatTrace` could be absorbed into the class since it's technically a degree 0 polynomial fit.

`FlatTrace` doesn't currently include any fitting functionality, but defines a flat trace at a specific pixel.  Similarly, `ArrayTrace` provides the flexibility to pass a custom array of values (again, no fitting).  This functionality could, in theory, all be merged into a single class, but then the parameter arguments become quite confusing with some irrelevant under certain circumstances and the internal logic checks get cumbersome.  Unless we want to abandon this functionality, I really don't think a single class with a single `__init__` is the way to go.

> There should be one user-facing tracing class named `Trace` to prevent the need to remember variant names like `KosmosTrace`, `PolyTrace`, etc.

As mentioned above, we have some current trace implementations that allow the user to _define_ a trace, and then this effort (and previously `KosmosTrace`) which _fits_ a trace to the data given a set of input options.  I think there needs to be a distinction between these, both for the user and within the implementation.

One option that might achieve the same rough goal of simplifying the list of classes ,while still preserving the necessary distinction would be to:
* merge `ArrayTrace` functionality into the base `Trace` class.  The default initializing of this default trace then becomes `Trace(trace_array)`.
* either have subclasses (`FlatTrace(30)`, `PolyTrace(guess=30, peak_method='gaussian', ...)`, `SpineTrace(...)`) _or_ classmethods (`Trace.at_pixel(30)`, `Trace.fit_polynomial(guess=30, peak_method='gaussian', ...)`, `Trace.fit_spline(...)`) which handle the various cases with their own independent API and docstrings.  These all effectively do some things under-the-hood (either building or fitting the array) and set the internal array.  The benefit of subclasses over classmethods is that we can make better use of the dataclass infrastructure used in specreduce to track the input parameters and allow the user to revise them and have the trace itself update.  Classmethods would effectively be a one-time use which _resolves_ to the array, forgetting about the input parameters used to create them.  If the user wants to modify the trace, they would have to manually edit the array, or start from scratch and build a new instance.  Either approach could easily share code for the fitting functionality while just swapping out the astropy model that is used.

> Given the chance to design tracing from scratch, they would not include spline fitting and the ability to bin.

I agree about the spline, but it also adds very little overhead to keep around as long as we have re-usable code.  But I have no objection to deprecating it if we don't think there is ever a use case for it over other options like polynomial.

Regarding binning, I agree that we should at least disable it by default, but I would imagine it could be useful if you have a particularly bright pixel away from the actual source that would otherwise pull the chosen `peak_method` away from the source.  There are likely better ways to handle this situation though (masking the input image, outlier detection, etc).  So maybe we can start deprecating binning but not actually remove it until we're sure we have something in place that can handle these situations (if they're a concern at all)?

camipacifici:
My 2 cents:
- the long list of classmethods under a single Trace is very tempting because it is the most intuitive (at least to me), but I see all the potential issues that come with it.
- I like the `FitTrace` option (better than AutoTrace to me) and its flexibility ("can then accept any generic astropy model").

I do not see problems with leaving FlatTrace and ArrayTrace alone if the difference with FitTrace is well explained in the docs. For example, an explanation of how a FlatTrace and a FitTrace with a polynomial of order 0 are different would be necessary.

PatrickOgle:
@kecnry Very useful discussion.  I agree with this:

"The other option which involves less breaking changes and that I've proposed in this PR, would be to leave FlatTrace and ArrayTrace alone and to rename KosmosTrace to AutoTrace (or FitTrace) and have the same classmethods or subclasses for various models (polynomial, etc) as I explained above. We would likely want to leave KosmosTrace as a deprecated alias for a little while to give users some time to migrate to the same name. The huge benefit to this, in my opinion, is that the "base" AutoTrace can then accept any generic astropy model and does not need to be limited to the ones that we explicitly expose as subclasses or classmethods. This also allows all "fitting" trace classes to have a method to refit the array whenever the parameters change, while not exposing this useless method to "define" trace classes."

I agree with @camipacifici that 'FitTrace' is better than 'AutoTrace'

Regarding this:
"Regarding binning, I agree that we should at least disable it by default, but I would imagine it could be useful if you have a particularly bright pixel away from the actual source that would otherwise pull the chosen peak_method away from the source. There are likely better ways to handle this situation though (masking the input image, outlier detection, etc). So maybe we can start deprecating binning but not actually remove it until we're sure we have something in place that can handle these situations (if they're a concern at all)?"

--KosmosTrace requires binning to avoid having an even crazier trace.  Polynomial Trace doesn't benefit as much from binning, but it is OK to have it as long as bin = 1 is the default.

specreduce/tracing.py
ojustino: I changed their representation in the docstring. It looks like consensus moved away from the subclass approach.

Passing in strings could be helpful, but I would put it below a few other issues for the time being... I'll keep this unresolved and reassess closer to merge.

specreduce/tracing.py
ojustino: Patrick wants the default equal to the number of dispersion columns. What do you think?

kecnry: Works for me.  We might want some logic to skip binning entirely for the case where it's equal to the image size or larger (or raise an error for the larger case).

CHANGES.rst
```diff
@@ -30,6 +30,14 @@ Bug Fixes
 - Fixed boxcar weighting to handle zero width and edge of image cases [#141]
 
 
+1.1.0
+-----
+
+New Features
+^^^^^^^^^^^^
+
+- Added ``PolyTrace`` class [#128]
```
kecnry, line 39: needs updating for `FitTrace` and to mention deprecation of `KosmosTrace`

specreduce/tracing.py
```diff
@@ -335,12 +346,135 @@ def __post_init__(self):
...
+    """
+    __doc__ += FitTrace.__doc__
+    pass
+
+
+@dataclass
+class PolyTrace(Trace):
+    """
+    Fit a trace to an image's peak cross-dispersion pixels according to
+    a polynomial function.
```
kecnry, line 378: will this class be removed entirely then in favor of `FitTrace(model=Polynomial1D(...), ...)`?

kecnry (commented):
Thanks for all the hard work... I think we're converging and can do a full code review once everyone is happy with the general direction.

I've updated https://github.com/spacetelescope/jdaviz/pull/1695 to keep up-to-date with the changes here if anyone wants to test how this works interactively within jdaviz.

specreduce/tracing.py
ojustino: Yes, I can dump it in an interactive rebase. I left it for now in case people still wanted to test it.

ojustino:
Following the discussion above, I replaced all mentions of `KosmosTrace` with `FitTrace` across the codebase and documentation. I also changed the default number of bins to `None`, which updates to the image's number of dispersion pixels if the user doesn't specify that argument. The gaussian `peak_method` is relatively slow on images with hundreds of dispersion pixels (or more) for reasons I mentioned earlier when comparing the performances of `FitTrace` and the standalone `PolyTrace` class.

(Since nothing has been finalized yet, I kept the separate `PolyTrace` class around in case anyone feels it needs another look. Detailed reviews should focus on `FitTrace`, and if it's sufficient, I'll delete `PolyTrace` before merging. I believe the coverage test will pass once it's gone.)

Questions:
1. If we don't want to feature cases that require binning, should we choose another image for `notebook_sandbox/jwst_boxcar/tracing_options.ipynb`? I can't get a good trace there unless I bin.
2. What should the default case be for `FitTrace`? I kept it as `peak_method='gaussian', trace_model=models.Polynomial(degree=3)` for now since that's what `KosmosTrace` had, but that's a spline fit and it seems we favor the polynomial route.

PatrickOgle:
FitTrace is not doing a good job of fitting the trace with either Polynomial1D or Chebyshev1D and peak_method = ‘gaussian’ or ‘max’:
trace_poly = FitTrace(img,100,trace_model=models.Chebyshev1D(2), peak_method='gaussian')
Screen Shot 2022-11-09 at 4 03 19 PM
(red = Flat Trace, magenta=FitTrace)

PatrickOgle:
1. It is OK to have binning in the examples to illustrate its use.
2. The suggested defaults are fine.  I am agnostic as to which model (Polynomial1D, Chebyshev, or Legendre) is default. I would pick Polynomial1D, just to be definitive. They all seem to work well for my test cases.

PatrickOgle (approved):
Tested new FitTrace class on both bright and faint 2D spectra.  It works well for all models, and is a great improvement on KosmosTrace, where the spline fit tends to wander in some cases. Extraction using resultant trace with BoxCarExtract appears to give accurate results. One lien for future consideration--the fitting is slow for large images using gaussian peak and no binning. This will be a hinderance for bulk processing of large numbers of spectra and also for visualization in Jdaviz.

specreduce/tracing.py
```diff
@@ -164,9 +162,9 @@ class KosmosTrace(Trace):
         direction is axis 1.
     bins : int, optional
         The number of bins in the dispersion (wavelength) direction
-        into which to divide the image. Use fewer if KosmosTrace is
-        having difficulty, such as with faint targets.
-        Minimum bin size is 4. [default: 20]
+        into which to divide the image. If not set, defaults to one bin
+        per dispersion (wavelength) pixel in the given image. If set,
+        requires a minimum of 4 bins. [default: None]
```
kecnry, line 167: is a lower limit of 4 bins sufficient or should we have a limit based on the order of the input model?

specreduce/tracing.py
```diff
@@ -221,17 +223,26 @@ def __post_init__(self):
         if self._disp_axis != 1:
             raise ValueError('dispersion axis must equal 1')
 
+        cols = img.shape[self._disp_axis]
+        if self.bins is None:
+            self.bins = cols
+            # self.bins = max(20, img.shape[self._disp_axis] / 4)  # which default?
```
kecnry, line 229:
```suggestion
```

kecnry (commented):
I am in favor of getting this in and we can always make some minor tweaks as it gets some actual usage.  I didn't go through all the sandbox notebooks in any detail, but the implementation in specviz2d seems to work quite well and behave as expected.  I'll hold off on official approval for now since I have a few small questions, but assuming those don't result in major changes will approve as soon as they're answered.

CHANGES.rst
```diff
@@ -4,9 +4,17 @@
...
+- The new FitTrace class (see "API Changes" below) introduces the ability to
+take a polynomial trace of an image [#128]
+
 API Changes
 ^^^^^^^^^^^
 
+- Renamed KosmosTrace as FitTrace, a conglomerate class for traces that are fit
+to images instead of predetermined [#128]
+- The default number of bins for FitTrace is now its associated image's number
+of dispersion pixels instead of 20. Its default peak_method is now 'max'. [#128]
```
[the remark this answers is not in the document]
ojustino: I favor giving `KosmosTrace` the new default for consistency's sake. Are there examples of `KosmosTrace` usage in the wild that need the old default? There's one in JDAT Notebooks, but that notebook is technically under review, so we can address the change as part of it.

kecnry (approved):
Thanks for all the iterations!  I can't think of anything else and this seems to behave well now with the exception of being slow for a large number of bins (which is not surprising).

specreduce/tracing.py
```diff
@@ -177,25 +176,29 @@ class KosmosTrace(Trace):
...
+        ``centroid``: Takes the centroid of the window within in bin.
+        ``max``: Saves the position with the maximum flux in each bin.
+        [default: ``max``]
     """
-    bins: int = 20
+    bins: int = None
     guess: float = None
     window: int = None
-    peak_method: str = 'gaussian'
+    trace_model: Model = field(default=models.Polynomial1D(degree=2))
```
tepickering, line 200: i would suggest setting the default degree to 1, i.e. a line. better to start with the assumption of no curvature and work up from there, if needed.

CHANGES.rst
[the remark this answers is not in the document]
tepickering: i don't think we need to worry about `KosmosTrace` usage in the wild so API breakage is fine.

tepickering (approved):
really, really nice work! this cleaner, more general, and more flexible implementation is a big improvement. i have one very minor suggestion and note that there are a couple of new lines missed by coverage. otherwise, it's good to go as far as i'm concerned. 👍

PR #134, opened 2022-09-08
MNT: Remove version template

pllim:
Do you really need it?

I wonder if this fixes the weird version generated for latest doc...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,17 +65,5 @@ if 'build_docs' in sys.argv or 'build_sphinx' in sys.argv:
     print(DOCS_HELP)
     sys.exit(1)
 
-VERSION_TEMPLATE = """
-# Note that we need to fall back to the hard-coded version if either
-# setuptools_scm can't be imported or setuptools_scm can't determine the
-# version, so we catch the generic 'Exception'.
-try:
-    from setuptools_scm import get_version
-    version = get_version(root='..', relative_to=__file__)
-except Exception:
-    version = '{version}'
-""".lstrip()
-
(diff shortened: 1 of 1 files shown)


pllim:
p.s. My other PR without this patch also reports the correct dev version. So I dunno 🤷

tepickering:
i wonder if there's confusion somewhere because `HEAD` and `1.1.0` are the same at the moment. we could alternatively use `pyproject.toml` to configure where `version.py` gets written. that's the more "modern" approach...

pllim:
Yes, you could. I heard that some projects got rid of `setup.py` entirely but you do need a much newer version of pip for that. If you are interested in that, please get in touch with @WilliamJamieson .

If you think this PR is irrelevant, feel free to close without merging. Thanks!

tepickering:
my last several personal projects are `setup.py`-free so i'm familiar with the approach. i haven't run into any issues with `pip`, but i also keep it diligently up-to-date. that all said, i think it's worth keeping it around here for the time being for the help messages and consistency with the rest of the `astropy` ecosystem. i think the changes here are a step in the right direction that simplifies `version.py` and keeps it inline with the `setuptools_scm` standard.

WilliamJamieson:
If you want to go with the pure pyproject.toml, you will want to specify setuptools >= 61 and setuptools_scm[toml] >= 3.4 in your build system. Those I think are the minimum versions for full pyproject.toml support.

tepickering:
i use 45 and 6.2 as min versions per the example at https://pypi.org/project/setuptools-scm/.

tepickering:
anyway, merging this now and further modernizing of the infrastructure can be handled in future PRs.

PR #135, opened 2022-09-08
DOC: No need to automodapi the top-level package

pllim:
This would remove the duplicate class API.


tepickering:
thanks for this! i'll just go ahead and merge this since it's a trivial fix.

PR #136, opened 2022-09-08
MNT: Remove dead code from APE 17

pllim:
I think this stuff is left over from astropy-helpers days.

Also see https://github.com/astropy/astropy-APEs/blob/main/APE17.rst


tepickering:
it is indeed `astropy-helpers` cruft that i hadn't gotten around to cleaning up. there's more to do to clean up the package infrastructure here to bring it more inline with modern standards, but this is a good start. thanks!

PR #146, opened 2022-10-13
BUG: fix passing single Trace object to Background

kecnry:
* `Background(image, trace, width)` used to expect `trace` to be an iterable and interpret it as a list of floats, each representing a trace (and therefore raise an error about window overlaps).  This syntax is now handled as expected: as a single background window, equivalent to `Background(image, [trace], width)`.


ojustino (approved):
Another good step on the path to robustness.

PR #147, opened 2022-10-25
Better management of Trace attribute mutability

### LINKED ISSUE #126: FlatTrace: possible points of confusion when editing existing object
For example, shifting edits the internal trace, but not `trace_pos`, but changes to `trace_pos` and `trace` do not affect each other:

```
from specreduce.utils.synth_data import make_2dspec_image
from specreduce.tracing import FlatTrace

img = make_2dspec_image()

trace = FlatTrace(img, 10)
print(trace.trace_pos, trace.trace[0])
# 10, 10

trace_shifted = trace + 10
print(trace_shifted.trace_pos, trace_shifted.trace[0])
# 10, 20

trace_shifted.trace_pos = 15
print(trace_shifted.trace_pos, trace_shifted.trace[0])
# 10, 20
```

Updating the trace_pos during a shift (or addition) can be done easily by overloading `Trace.shift`.  Do we also want to implement a setter on the `trace_pos` attribute to update `trace.trace`?  What about vice-versa?  Or do we make `trace.trace` read-only and require any edits to be by setting `trace_pos` or shifting?

bmorris3:
#126 points out that you can edit attributes of the `Trace` class in such a way that they are no longer self-consistent with each other.

In this PR, I've found a way to recast the `*Trace`s as subclasses of a new `BaseTrace` (perhaps there's a better name?). These classes all use `@dataclass(frozen=True)` which enforces that their attributes are immutable for the user-facing API. Under the hood, there's a slightly ugly workaround which allows us to (re)set these "read-only" attributes via the public-facing methods like `Trace.shift`, for example. 

Happy to discuss as needed!

Fixes #126

### Tiny demo

```python
import pytest
from specreduce.utils.synth_data import make_2dspec_image
from specreduce.tracing import FlatTrace

img = make_2dspec_image()

trace = FlatTrace(img, 10)
print(trace.trace_pos, trace.trace[0])
# used to produce: 10, 10
# now produces: 10, 10

trace_shifted = trace + 10
print(trace_shifted.trace_pos, trace_shifted.trace[0])
# used to produce: 10, 20
# now produces: 20, 20

with pytest.raises(AttributeError):
    # this didn't raise an error before, but does now
    trace_shifted.trace_pos = 15
```

### DIFF
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -8,7 +8,7 @@ from astropy.nddata import NDData
 from astropy import units as u
 
 from specreduce.extract import _ap_weight_image, _to_spectrum1d_pixels
-from specreduce.tracing import FlatTrace, BaseTrace
+from specreduce.tracing import FlatTrace, Trace
 
 __all__ = ['Background']
 
@@ -77,7 +77,7 @@ class Background:
             cross-dispersion axis
         """
         def _to_trace(trace):
-            if not isinstance(trace, BaseTrace):
+            if not isinstance(trace, Trace):
                 trace = FlatTrace(self.image, trace)
 
             # TODO: this check can be removed if/when implemented as a check in FlatTrace
@@ -93,7 +93,7 @@ class Background:
             self.bkg_array = np.zeros(self.image.shape[self.disp_axis])
             return
 
-        if isinstance(self.traces, BaseTrace):
+        if isinstance(self.traces, Trace):
             self.traces = [self.traces]
 
         bkg_wimage = np.zeros_like(self.image, dtype=np.float64)
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -8,7 +8,7 @@ from astropy.nddata import NDData
 from astropy import units as u
 
 from specreduce.extract import _ap_weight_image, _to_spectrum1d_pixels
-from specreduce.tracing import Trace, FlatTrace
+from specreduce.tracing import FlatTrace, BaseTrace
 
 __all__ = ['Background']
 
@@ -77,7 +77,7 @@ class Background:
             cross-dispersion axis
         """
         def _to_trace(trace):
-            if not isinstance(trace, Trace):
+            if not isinstance(trace, BaseTrace):
                 trace = FlatTrace(self.image, trace)
 
             # TODO: this check can be removed if/when implemented as a check in FlatTrace
@@ -93,7 +93,7 @@ class Background:
             self.bkg_array = np.zeros(self.image.shape[self.disp_axis])
             return
 
-        if isinstance(self.traces, Trace):
+        if isinstance(self.traces, BaseTrace):
             self.traces = [self.traces]
 
         bkg_wimage = np.zeros_like(self.image, dtype=np.float64)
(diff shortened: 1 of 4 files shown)


tepickering:
looks pretty good! i need to take a deeper dive, but one substantive comment is that `trace_pos` is only really meaningful for `FlatTrace` and shouldn't be in the base class. in fact, it should just be an `InitVar` for `FlatTrace` that is used to construct `self._trace` and not become an attribute. if we keep it around as a property, it should get the position directly from `self._trace`.

kecnry:
This is a clever workaround, and in general I like it.   But before moving forward in this direction, I think we need to open a wider discussion over whether we want to support the user setting _any_ attributes across _any_ of the classes (background and extraction included) and expect the internal information in those objects to update respectively.  Currently I think the user can set any of those attributes, but similar to the cases here, internal arrays aren't updated accordingly, so I think we can decide to go either direction as long as we do so consistently across the board.  But if we do want some/most to be editable, then would we need to set custom setters for each case or would that make this approach unfeasible?

As a side note, correct me if I'm wrong @tepickering, but I think `Trace` already was the base class and wasn't meant to be directly by the user... so maybe we wouldn't need separate `BaseTrace` and `Trace` classes here.

pllim:
I don't see any test. Shouldn't there be tests to ensure this patch fixes the problem?

bmorris3:
In the last two commits I've:
* removed `BaseTrace` and implemented the same functionality on the `Trace` class, thanks @kecnry ([comment](https://github.com/astropy/specreduce/pull/147#issuecomment-1291965589))
* moved the `trace_pos` attribute to be specific to only the `FlatTrace` subclass, thanks @tepickering ([comment](https://github.com/astropy/specreduce/pull/147#issuecomment-1291283353))
* added tests, thanks @pllim ([comment](https://github.com/astropy/specreduce/pull/147#issuecomment-1302218988))

specreduce/background.py
```diff
@@ -8,7 +8,7 @@
 from astropy import units as u
 
 from specreduce.extract import _ap_weight_image, _to_spectrum1d_pixels
-from specreduce.tracing import Trace, FlatTrace
+from specreduce.tracing import FlatTrace, Trace
```
ojustino, line 11: Is this just for alphabetization purposes?

specreduce/tracing.py
```diff
@@ -92,10 +109,12 @@ class FlatTrace(Trace):
     trace_pos : float
         Position of the trace
     """
-    trace_pos: float
+    _trace_pos: (float, np.ndarray) = field(repr=False)
```
ojustino, line 112: Can you help me understand when/why this `FlatTrace` argument would take an array?

ojustino (commented):
Thanks for your work, @bmorris3. I found this approach a little hard to follow at first. My understanding now is it seems it has to be a backdoor-style fix due to some limitations of `dataclass`, so I can get behind it.

I think it makes sense for most (if not all) of our current selection of attributes to be read-only, so I support an approach like the one here over the alternative of allowing editable attributes and needing to update internal arrays. I feel the same about the non-trace operations, too, but that work would be for another ticket.

I agree that we wouldn't need both a `BaseTrace` and `Trace`, so it's good we're sticking with `Trace` as the base class.

hpparvi:
Hi @bmorris3. Do you remember if there was any off-GitHub discussion that led to this PR not being merged? I like the idea of making the tracing classes immutable and would be happy to help get this merged, though probably only after @cshanahan1's masking work.

PR #148, opened 2022-11-09
Introducing spectral extractions for non-flat traces

### LINKED ISSUE #102: HorneExtract treatment of non-flat trace
HorneExtract computes the profile by collapsing the spectrum along the dispersion-axis, and then applying that same profile across each "column" from the trace.  This will likely result in an artificially broad profile if the trace is not strictly flat.  For non-flat traces, we could shift each column to align the traces or rotate and interpolate before collapsing.  In the meantime, we should require a flat trace or raise an error/warning. (cc @orifox)

bmorris3:
### Description 

This PR implements Horne spectral extraction with non-flat traces. For a demo, see [this example notebook](https://gist.github.com/bmorris3/428a0b012305fd6742c944f2c3c11a04).

specreduce Horne extractions currently support flat traces. In the flat trace case, the sum of the spectrum in the dispersion dimension produces a flux profile in the spatial dimension, which we fit with a (1D) Gaussian. The model profile is used to construct a kernel, which applies weights to each pixel in the extracted 1D spectrum.

For a non-flat trace, the profile in the will be "smeared" when summed in the dispersion dimension, which is not accurately approximated by a Gaussian. Less flat traces will produce poorer kernels, and less accurate spectral extractions. 

The trick used here to support non-flat traces is to roll the pixels along the spatial dimension until the trace is flat, and then perform the extraction on the shifted image as usual. There are no actual changes to the Horne implementation, just a manipulation of the image in the case of a non-flat trace to "make it flat".

The trick within the trick is that we can use fancy indexing to flatten the trace without interpolation or resampling, and do it efficiently. In this PR, we only re-index the underlying image. 

My strategy here is to (1) never make up data (no interp/sampling business), and (2) keep the diff small. 

### Possible Improvements
* We'll need some tests and documentation, of course. (I'm posting this draft PR to get initial feedback.)
* @kecnry pointed out a possible (small?) improvement. This PR casts floats in the trace to integers for rolling the pixels along the spatial dimension (without interpolation). That means we can expect a small spatial difference between the center of the spatial profile and the intended center of the new flat trace. We could pass along that small difference to the initial guess for the mean in the Gaussian 1D spatial profile. This may not affect the results much, because the rolls are accurate to better than a pixel, and the spatial width of the observed trace is usually a few pixels -- so the initial guess is usually good enough.

Closes #102

[🐱](https://jira.stsci.edu/browse/JDAT-2389)

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -117,6 +117,34 @@ def _ap_weight_image(trace, width, disp_axis, crossdisp_axis, image_shape):
     return wimage
 
 
+def _align_along_trace(img, array_trace, disp_axis=1, crossdisp_axis=0):
+    """
+    Given an arbitrary trace ``array_trace`` (an np.ndarray), roll
+    all columns of ``nddata`` to shift the NDData's pixels nearest
+    to the trace to the center of the spatial dimension of the
+    NDData.
+    """
+    # TODO: this workflow does not support extraction for >2D spectra
+    if not (disp_axis == 1 and crossdisp_axis == 0):
+        # take the transpose to ensure the rows are the cross-disp axis:
+        img = img.T
+
+    n_rows = img.shape[0]
+
+    # indices of all columns, in their original order
+    rows = np.broadcast_to(np.arange(img.shape[0])[:, None], img.shape)
+    cols = np.broadcast_to(np.arange(img.shape[1]), img.shape)
+
+    # we want to "roll" each column so that the trace sits in
+    # the central row of the final image
+    shifts = array_trace.astype(int) - n_rows // 2
+
+    # we wrap the indices so we don't index out of bounds
+    shifted_rows = np.mod(rows + shifts[None, :], img.shape[0])
+
+    return img[shifted_rows, cols]
+
+
 @dataclass
 class BoxcarExtract(SpecreduceOperation):
     """
@@ -462,6 +490,23 @@ class HorneExtract(SpecreduceOperation):
         img = np.ma.masked_array(self.image.data, or_mask)
         mask = img.mask
 
+        # If the trace is not flat, shift the rows in each column
+        # so the image is aligned along the trace:
+        if isinstance(trace_object, FlatTrace):
+            mean_init_guess = trace_object.trace
+        else:
+            img = _align_along_trace(
+                img,
+                trace_object.trace,
+                disp_axis=disp_axis,
+                crossdisp_axis=crossdisp_axis
+            )
+            # Choose the initial guess for the mean of
+            # the Gaussian profile:
+            mean_init_guess = np.broadcast_to(
+                img.shape[crossdisp_axis] // 2, img.shape[disp_axis]
+            )
+
         # co-add signal in each image column
         ncols = img.shape[crossdisp_axis]
         xd_pixels = np.arange(ncols)  # y plot dir / x spec dir
@@ -483,9 +528,9 @@ class HorneExtract(SpecreduceOperation):
         norms = []
         for col_pix in range(img.shape[disp_axis]):
             # set gaussian model's mean as column's corresponding trace value
-            fit_ext_kernel.mean_0 = trace_object.trace[col_pix]
-            # NOTE: support for variable FWHMs forthcoming and would be here
 
+            # NOTE: support for variable FWHMs forthcoming and would be here
+            fit_ext_kernel.mean_0 = mean_init_guess[col_pix]
             # fit compound model to column
             fitted_col = fit_ext_kernel(xd_pixels)
 
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -117,9 +117,9 @@ def _ap_weight_image(trace, width, disp_axis, crossdisp_axis, image_shape):
     return wimage
 
 
-def _align_along_trace(img, array_trace, disp_axis=1, crossdisp_axis=0):
+def _align_along_trace(img, trace_array, disp_axis=1, crossdisp_axis=0):
     """
-    Given an arbitrary trace ``array_trace`` (an np.ndarray), roll
+    Given an arbitrary trace ``trace_array`` (an np.ndarray), roll
     all columns of ``nddata`` to shift the NDData's pixels nearest
     to the trace to the center of the spatial dimension of the
     NDData.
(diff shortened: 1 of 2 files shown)


specreduce/extract.py
```diff
@@ -404,6 +432,23 @@ def __call__(self, image=None, trace_object=None,
...
+                img,
+                trace_object.trace,
+                disp_axis=disp_axis,
+                crossdisp_axis=crossdisp_axis
+            )
+            # Choose the initial guess for the mean of
+            # the Gaussian profile:
+            mean_init_guess = np.broadcast_to(
+                img.shape[crossdisp_axis] // 2, img.shape[disp_axis]
+            )
```
kecnry, line 450: would just swapping this out with the "remainder" from casting `trace_object.trace` to their integer rolls be enough to correct the means to the actual trace position?

specreduce/extract.py
```diff
@@ -123,6 +123,34 @@ def _to_spectrum1d_pixels(fluxes):
...
+    cols = np.broadcast_to(np.arange(img.shape[1]), img.shape)
+
+    # we want to "roll" each column so that the trace sits in
+    # the central row of the final image
+    shifts = array_trace.astype(int) - n_rows // 2
+
+    # we wrap the indices so we don't index out of bounds
+    shifted_rows = np.mod(rows + shifts[None, :], img.shape[0])
+
+    return img[shifted_rows, cols]
```
kecnry, line 151: is this cheaper or different than `np.roll`/`rollaxis`?

bmorris3: It's different because [np.roll](https://numpy.org/doc/stable/reference/generated/numpy.roll.html) is written to support rolling an array over one axis by a constant amount everywhere on that axis (their "shift" parameter takes one integer shift per rolled axis).

specreduce/extract.py
bmorris3: I think it would improve the initial guess for the mean parameter given to the Gaussian spatial profile fit by <1 pixel.

specreduce/extract.py
kecnry: According to the docs, `shift` can be a tuple (one integer shift per row), is that not what you need?

bmorris3: See my comment above – shift takes one integer shift per *axis*.

kecnry: Ah, I misread the docs as `shift` being the same length as the axis corresponding to `axis`, not the same length as the list of axes passed to `axis`.  I guess you'd have to iterate manually then and rebuild the whole image, which probably would be more expensive than your manual implementation.  Thanks for the explanation.

specreduce/extract.py
```diff
@@ -123,6 +123,34 @@ def _to_spectrum1d_pixels(fluxes):
                       flux=fluxes)
 
 
+def _align_along_trace(img, array_trace, disp_axis=1, crossdisp_axis=0):
```
ojustino, line 126: What do you think of changing `array_trace` to `trace_array` (or something else) to avoid confusion with or the connotation that it comes from an `ArrayTrace` object?

specreduce/extract.py
```diff
@@ -123,6 +123,34 @@ def _to_spectrum1d_pixels(fluxes):
...
+    # TODO: this workflow does not support extraction for >2D spectra
+    if not (disp_axis == 1 and crossdisp_axis == 0):
+        # take the transpose to ensure the rows are the cross-disp axis:
+        img = img.T
+
+    n_rows = img.shape[0]
+
+    # indices of all columns, in their original order
+    rows = np.broadcast_to(np.arange(img.shape[0])[:, None], img.shape)
+    cols = np.broadcast_to(np.arange(img.shape[1]), img.shape)
```
ojustino, line 142:
```suggestion
    rows = np.broadcast_to(np.arange(n_rows)[:, None], img.shape)
    cols = np.broadcast_to(np.arange(img.shape[disp_axis]), img.shape)
```

specreduce/extract.py
```diff
@@ -123,6 +123,34 @@ def _to_spectrum1d_pixels(fluxes):
...
+
+    # we want to "roll" each column so that the trace sits in
+    # the central row of the final image
+    shifts = array_trace.astype(int) - n_rows // 2
+
+    # we wrap the indices so we don't index out of bounds
+    shifted_rows = np.mod(rows + shifts[None, :], img.shape[0])
```
ojustino, line 149:
```suggestion
    shifted_rows = np.mod(rows + shifts[None, :], n_rows)
```

specreduce/extract.py
```diff
@@ -425,9 +470,9 @@ def __call__(self, image=None, trace_object=None,
         norms = []
         for col_pix in range(img.shape[disp_axis]):
             # set gaussian model's mean as column's corresponding trace value
-            fit_ext_kernel.mean_0 = trace_object.trace[col_pix]
-            # NOTE: support for variable FWHMs forthcoming and would be here
 
+            # NOTE: support for variable FWHMs forthcoming and would be here
+            fit_ext_kernel.mean_0 = mean_init_guess[col_pix]
```
ojustino, line 475: The order of the code and comment were intentional here. Could you swap them back?

specreduce/extract.py
```diff
@@ -123,6 +123,34 @@ def _to_spectrum1d_pixels(fluxes):
...
+    all columns of ``nddata`` to shift the NDData's pixels nearest
+    to the trace to the center of the spatial dimension of the
+    NDData.
+    """
+    # TODO: this workflow does not support extraction for >2D spectra
+    if not (disp_axis == 1 and crossdisp_axis == 0):
+        # take the transpose to ensure the rows are the cross-disp axis:
+        img = img.T
+
+    n_rows = img.shape[0]
```
ojustino, line 138:
```suggestion
    n_rows = img.shape[crossdisp_axis]
```
You could also add `n_cols` so operations later on look more consistent.

ojustino (commented):
This is a creative approach that seems to work upon initial investigation. I would like to test the extractions more thoroughly, but I have some comments on the code in the meantime.

specreduce/extract.py
bmorris3: Sure!

specreduce/extract.py
bmorris3: My mistake!

specreduce/extract.py
bmorris3: I've changed this expression in 078b48d.

specreduce/extract.py
bmorris3: Changed in 078b48d.

kecnry:
I think this is ready for a final code-review once test coverage is added.

bmorris3:
The remaining failure comes from astropy convolution, see https://github.com/astropy/astropy/issues/14025

ojustino:
It looks like the big difference since my last check is the test, which seems reasonable.

I put more thought into the resampling procedure and am wondering if it's problematic in cases where it's not actually needed because the trace is truly flat. The following is the same as cell 4 of your gist's notebook except it skips the transformation step:
```
image_nddata = NDData(data, uncertainty=error)
flat_trace = FlatTrace(data, 30)
spec1d_init = HorneExtract(image_nddata, trace)()

fig, ax = plot_result(image_nddata, flat_trace, spec1d_init)
```
Is the difference in the results a cause for concern?

bmorris3:
Hi @ojustino, thanks for your investigation and this comment. The "untransformed" image should indeed produce a distinct spectral extraction from the one I made in the example notebook which does transform `image_nddata` via `transform_flat_to_custom_trace`. That function does some interpolation and resampling of the "original" image which takes advantage of interpolation and resampling via `scipy.ndimage.zoom`. This method is not flux conserving and not written in a way so that the "untransformed" image matches the "transformed" one when the trace is flat. I only wrote that method to fudge non-flat traces from a real and flat example.

bmorris3:
I've reverted the very last commit. The reverted commit pinned the setuptools version, due to an upstream astropy failure related to the convolution module. The astropy fix was merged in astropy/astropy#14035, so we no longer need the pin here.

ojustino (approved):
@bmorris3, that makes sense, thanks. Looks like I was fixating on the example instead of the actual code. The code looks good to me and the test makes sense, so I approve from that perspective. I would be interested in seeing feedback from others with more experience in extraction using non-flat traces, but I have no objections based on the ones I've tried.

kecnry (approved):
I agree - let's get this in and make any small tweaks once there is some user feedback!

bmorris3:
Thanks both for the reviews. 🎉

PR #151, opened 2022-11-30
BUG: Handle invalid masked_invalid mask

pllim:
I came across this while investigating https://github.com/spacetelescope/jdaviz/issues/1877 . `np.ma.masked_invalid(img).mask` could return `False`. This is not being recognized by `specutils`, which will then emit a confusing message about the mask having mismatch shape `()`.

Possible suspects:

* https://github.com/numpy/numpy/pull/22046
* https://github.com/numpy/numpy/pull/22406

Also see:

* #153

### DIFF
--- a/specreduce/core.py
+++ b/specreduce/core.py
@@ -64,7 +64,7 @@ class _ImageParser:
             mask = image.mask
         else:
             mask = np.ma.masked_invalid(img).mask
-            if not mask:  # Could be False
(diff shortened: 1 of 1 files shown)


tepickering:
looks like this change needs some test coverage to made codecov happy...

pllim:
I don't know how to write that test. Suggestion welcome!

pllim:
I folded this PR to #153

PR #153, opened 2022-12-02
TST: Add dev numpy to devdeps, fix compat with numpy 1.24

pllim:
This might smoke out the error that #151 is trying to fix.

I also see this error in https://github.com/astropy/astropy-integration-testing/actions/runs/3604333353/jobs/6073585087

Possible suspects:

* https://github.com/numpy/numpy/pull/22046
* https://github.com/numpy/numpy/pull/22406

### DIFF
--- a/specreduce/core.py
+++ b/specreduce/core.py
@@ -75,6 +75,9 @@ class _ImageParser:
         spectral_axis = getattr(image, 'spectral_axis',
                                 np.arange(img.shape[disp_axis]) * u.pix)
 
+        if mask.ndim == np.ma.nomask:
+            mask = np.zeros(img.shape, dtype=bool)  # 0 = good
+
         return Spectrum1D(img * unit, spectral_axis=spectral_axis,
                           uncertainty=uncertainty, mask=mask)
 
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -359,6 +359,9 @@ class HorneExtract(SpecreduceOperation):
         else:
             mask = np.ma.masked_invalid(img).mask
 
+        if mask.ndim == np.ma.nomask:
+            mask = np.zeros(img.shape, dtype=bool)  # 0 = good
+
         if img.shape != mask.shape:
             raise ValueError('image and mask shapes must match.')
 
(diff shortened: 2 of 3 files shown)


pllim:
Hmm. Maybe #151 isn't the correct solution after all.

ojustino:
Hi @pllim, after some local testing, I think the best solution would be to drop `masked_where()` entirely. In the place that causes the error you found, all we really need is the mask; the other attributes that come with a `masked_array` are superfluous. All they do to create the mask is call `isfinite()`, so we can just do the same here.

This usage of `masked_where()` appears a few places in `specreduce` and I believe they should all should be changed in the same effort. That could be this one, or if you like, I can try it in a different pull request.

pllim:
If you want to change mask logic, it is up to you as I am not familiar with the algorithms here. I just did track down why numpy dev is giving error and it might or might not be upstream bug. I am going to ask numpy people.

specreduce/core.py
```diff
@@ -75,6 +75,9 @@ def _parse_image(self, image, disp_axis=1):
         spectral_axis = getattr(image, 'spectral_axis',
                                 np.arange(img.shape[disp_axis]) * u.pix)
 
+        if mask.ndim == 0:
```
pllim, line 78: This makes the tests pass but I am not actually sure if this is correct.

pllim:
As for the coverage, it will fix itself when numpy is released and your coverage job picks it up. Obviously it is covered because tests fail with dev numpy.

pllim:
Maybe we can merge by the end of the week if numpy makes it clear that they have no intentional of backing out this behavior.

ojustino:
@pllim I opened #155 to make the edits I talked about earlier since most of the places that trigger this issue don't actually need to use `mask_invalid()`. Do you have any objections to that approach?

ojustino:
The other PR should probably still integrate the changes you made to `tox.ini` in order to catch these errors more quickly in the future.

pllim:
@ojustino , yes, please do cherry-pick the dev numpy job over. If #155 supersedes this PR, feel free to close this without merge. Thanks!

PR #154, opened 2022-12-05
Update image handling in notebooks

ojustino:
These changes fix a few statements in more recent notebooks that needed adjustment after the changes to image handling in #144.


kecnry (approved):
Good catch, thanks!

PR #155, opened 2022-12-05
Use simpler checks for invalid values

ojustino:
In response to findings from #153, I've changed most uses of `np.ma.mask_invalid()` to `~np.isfinite()` since we only sought a mask instead of an entire masked array object most times it was used. We do still want a masked array as the final result of `FitTrace`, so it remains in use there.

These changes should help tests pass with numpy v1.24.0rc1. Would close #153.

### DIFF
--- a/specreduce/tests/test_tracing.py
+++ b/specreduce/tests/test_tracing.py
@@ -125,7 +125,7 @@ def test_fit_trace():
         FitTrace(img, bins=20., trace_model=models.Polynomial1D(2))
 
     # ensure non-equipped models are rejected
-    with pytest.raises(ValueError, match=r'trace_model must be one of*'):
+    with pytest.raises(ValueError, match=r'trace_model must be one of'):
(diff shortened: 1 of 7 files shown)


specreduce/tests/test_tracing.py
```diff
@@ -144,15 +144,15 @@ def test_fit_trace():
...
         raise RuntimeError('Trace was erroneously calculated on all-NaN window')
 
     # error on trace of all-nan image
     try:
         FitTrace(img_all_nans)
     except ValueError as e:
-        print(f"All-NaN image error message: {e}")
+        print(f"(expected) All-NaN image error message: {e}")
     else:
         raise RuntimeError('Trace was erroneously calculated on all-NaN image')
```
kecnry, line 157: would this test be simplified by using `pytest.raises`?

ojustino: Yes, it would. I just committed an update.

kecnry (approved):

Issue #158, opened 2022-12-14
NaNs outside background window still propagate into 1D spectra


ojustino:
From @kecnry:
> a single nan value in a 2d spectrum currently gets propagated to the entire column during background/background subtraction even if that pixel is outside the window used to create the background.  This then results in a nan in the resulting 1d background and 1d extracted spectrum and causes Horne extraction to fail entirely.

> To reproduce in specviz2d (try changing the windows in the spectral extraction plugin and exporting the background/background-subtracted image):

```
import numpy as np
from jdaviz import Specviz2d

from astropy.utils.data import download_file
fn = download_file('https://stsci.box.com/shared/static/exnkul627fcuhy5akf2gswytud5tazmw.fits', cache=True)
specviz2d = Specviz2d()
specviz2d.load_data(fn)
sp = specviz2d.app.get_data_from_viewer('spectrum-2d-viewer', 'Spectrum 2D')

sp.flux[30, 358] = np.nan

specviz2d = Specviz2d()
specviz2d.load_data(sp)
specviz2d.show()
```

This behavior comes back to issues the `speceduce` operations have with handling NaNs. The fixes for `Background`/`BoxcarExtract` and `HorneExtract` will happen in separate pull requests.

PR #159, opened 2022-12-14
Prevent (most) NaN propagation in `Background` and `BoxcarExtract`

ojustino:
This pull request should be part 1 of a multi-pronged response to #158. Tests are forthcoming.

### `Background`

Goal: NaNs outside the user-specified window should not propagate into the 1D spectrum of the background or the background-subtracted image. NaNs inside the user-specified window should be ignored.

Approach:
- Adjusted `_bkg_array`:
  - Take the user-provided mask (if it exists) into account by bringing in the `or_mask` approach from `HorneExtract` that also masks non-finite pixels.
  - Fixed a possible mistake in the `statistic='median'` case that may have previously allowed zero-weighted pixels to still make it into the median calculation.
- `bkg_image()` and `sub_image()` don't appear to need changes as long as `_bkg_array` handles NaNs appropriately.
- `bkg_spectrum()` and `sub_spectrum()` just need `np.sum` changed to `np.nansum`.

### `BoxcarExtract`

Goal: As with `Background`, NaNs outside the user-specified window should not propagate into the extracted 1D spectrum. _Unlike with `Background`_, NaNs inside the user-specified window should propagate to the extracted 1D spectrum. Zeroing them out of the sum could lure less careful users into finding spectral features that don't actually exist.

Approach:
- Assign no weight to non-finite pixels with zero weight while summing each column. This case needs to be handled specially because `np.nan * 0 = np.nan`, not 0. `_parse_image()` can't help because, for now, it only either adopts the user's mask or, if none is provided, masks every non-finite pixel. It can't only mask non-finite pixels outside the window.
  - The better approach going forward may be for `_parse_image()` to always defer to the user's mask, even if none is provided. Making our operations more resistant to unmasked NaNs we may see as a result would require major changes in another ticket.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,7 @@ Bug Fixes
 - Output 1D spectra from Background no longer include NaNs. Output 1D
 spectra from BoxcarExtract no longer include NaNs when none are present
 in the extraction window. NaNs in the window will still propagate to
-Boxcar1D's extracted 1D spectrum. [#159]
+BoxcarExtract's extracted 1D spectrum. [#159]
 
 - Backgrounds using median statistic properly ignore zero-weighted pixels
 [#159]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ API Changes
 Bug Fixes
 ^^^^^^^^^
 
+- Output 1D spectra from Background and BoxcarExtract no longer propagate
+NaNs in source images that contain them as long as none are present in the
+specified background window [#159]
+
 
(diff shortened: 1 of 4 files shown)


kecnry (commented):
Acts as I'd expect in the cases that used to result in a column of nans in the background image, and the implementation seems reasonable and clean to me!  Happy to re-review/approve once test coverage is added and passes.  Thanks!

tepickering (approved):
looks good to me, too. thanks for this! approved pending the added test coverage `codecov` is requesting.

PR #160, opened 2022-12-14
Update passenv syntax for tox 4.0.8

ojustino:
Reacting to an upstream update by `tox`, as in spacetelescope/jdaviz#1916.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,7 @@ isolated_build = true
 [testenv]
 
 # Pass through the following environment variables which may be needed for the CI
-passenv = HOME WINDIR LC_ALL LC_CTYPE CC CI
+passenv = HOME,WINDIR,LC_ALL,LC_CTYPE,CC,CI
 
 setenv =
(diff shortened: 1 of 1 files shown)


kecnry (approved):

tepickering:
thanks! this change has bitten a lot of projects. there's a weird failure in the `oldestdeps` test. might be spurious or it might need a change in lowest supported version of something (likely `numpy`).

ojustino:
I think the problem may have been that `oldestdeps` tests `specutils` all the way back to v1.7.0 even though we require v1.9.1 as of our last release. After updating the `oldestdeps` pin to v1.9.1 and resolving a few other version conflicts, I don't see the failure locally. Let's see how GitHub likes it...

kecnry:
Do any of those require changes to the pins in `setup.cfg`?  Or will `specutils>=1.9.1` already require at least the min versions of astropy/numpy in the CI matrix?

ojustino:
`specutils>=1.9.1` covers us with `astropy>=5.1`, which covers us with `numpy>=1.20`. I'm not sure about `matplotlib`, which is only in the "recommended" section of Astropy's setup file, but that shouldn't matter for the CI matrix, right?

PR #162, opened 2023-01-26
Implement 1D wavelength calibration classes

rosteen:
This implements the initial work needed for 1D wavelength calibration, defaulting to a basic linear fit to the input wavelength/pixel pairs. Does not implement autoidentification of lines or input of a preset named line list. I have a little bit of refinement left to do (e.g., making sure the internals are consistent on whether pixel values are stored as `Quantities` with `u.pix`) and need to add tests, but I figured it's to a point where I should get other eyes on it to make sure I'm fulfilling the consensus reached in the discussion on #152.

The following code snippet is what I was using to test this:

```
import astropy.units as u
import numpy as np
from specutils import Spectrum1D
from wavelength_calibration import WavelengthCalibration1D

flux = np.random.random(50)*u.Jy
sa = np.arange(0,50)*u.pix
spec = Spectrum1D(flux, spectral_axis = sa)
test = WavelengthCalibration1D(spec, [(0,5000*u.AA),(10,5100*u.AA),(20,5198*u.AA),(30,5305*u.AA)])
spec2 = test.apply_to_spectrum(spec)
spec2
```

### DIFF
--- a/docs/wavelength_calibration.rst
+++ b/docs/wavelength_calibration.rst
@@ -17,20 +17,20 @@ spectrum, you can currently use ``specreduce`` to:
 The `~specreduce.wavelength_calibration.WavelengthCalibration1D` class can be used
 to fit a dispersion model to a list of line positions and wavelengths. Future development
 will implement catalogs of known lamp spectra for use in matching observed lines. In the
-example below, the line positions (``line_list``) have already been extracted from
+example below, the line positions (``pixel_centers``) have already been extracted from
 ``lamp_spectrum``::
 
     import astropy.units as u
 	 from specreduce import WavelengthCalibration1D
-	 line_list = [10, 22, 31, 43]
+	 pixel_centers = [10, 22, 31, 43]
 	 wavelengths = [5340, 5410, 5476, 5543]*u.AA
-	 test_cal = WavelengthCalibration1D(lamp_spectrum, line_list,
-												  line_wavelengths=wavelengths)
+	 test_cal = WavelengthCalibration1D(lamp_spectrum, line_pixels=pixel_centers,
+										line_wavelengths=wavelengths)
     calibrated_spectrum = test_cal.apply_to_spectrum(science_spectrum)
 
 The example above uses the default model (`~astropy.modeling.functional_models.Linear1D`)
 to fit the input spectral lines, and then applies the calculated WCS solution to a second
-spectrum (``science_spectrum``). Any other ``astropy`` model can be provided as the
+spectrum (``science_spectrum``). Any other 1D ``astropy`` model can be provided as the
 input ``model`` parameter to the `~specreduce.wavelength_calibration.WavelengthCalibration1D`.
 In the above example, the model fit and WCS construction is all done as part of the
 ``apply_to_spectrum()`` call, but you could also access the `~gwcs.wcs.WCS` object itself
@@ -43,10 +43,11 @@ or ``input_spectrum`` properties are updated, since these will alter the calcula
 fit.
 
 You can also provide the input pixel locations and wavelengths of the lines as an
-`~astropy.table.QTable`, with columns ``pixel_center`` and ``wavelength``::
+`~astropy.table.QTable` with (at minimum) columns ``pixel_center`` and ``wavelength``,
+using the ``matched_line_list`` input argument::
 
     from astropy.table import QTable
     pixels = [10, 20, 30, 40]*u.pix
     wavelength = [5340, 5410, 5476, 5543]*u.AA
     line_list = QTable([pixels, wavelength], names=["pixel_center", "wavelength"])
-    test_cal = WavelengthCalibration1D(lamp_spectrum, line_list)
+    test_cal = WavelengthCalibration1D(lamp_spectrum, matched_line_list=line_list)
--- a/docs/wavelength_calibration.rst
+++ b/docs/wavelength_calibration.rst
@@ -1,0 +1,71 @@
+.. _wavelength_calibration:
+
+Wavelength Calibration
+======================
+
+Wavelength calibration is currently supported for 1D spectra. Given a list of spectral
+lines with known wavelengths and estimated pixel positions on an input calibration
+spectrum, you can currently use ``specreduce`` to:
+
+#. Refine the pixel position estimates of the lines based on features in the input
+   calibration spectrum.
+#. Fit an ``astropy`` model to the wavelength/pixel pairs to generate a spectral WCS
+   solution for the dispersion.
+#. Apply the generated spectral WCS to other `~specutils.Spectrum1D` objects.
+
+Calibration Lines
+-----------------
+
+``specreduce`` provides a `~specreduce.wavelength_calibration.CalibrationLine` class for
+encoding the information about each spectral line to be used in the dispersion model
+solution fitting. The minimum information required is the wavelength and estimated pixel
+location on the calibration spectrum of each line. By default, no refinement of the pixel
+position is done; to enable this feature you must provide a refinement method and any
+associated keywords (currently ``range`` defines the number of pixels on either side
+of the estimated location to use for all available refinement options). For example,
+to define a line and use ``specreduce`` to update the pixel location the the centroid
+of a gaussian fit to the line, you would do the following::
+
+	import astropy.units as u
+	from specreduce import CalibrationLine
+	test_line = CalibrationLine(calibration_spectrum, 6562.81*u.AA, 500,
+							   refinement_method="gaussian",
+							   refinement_kwargs={"range": 10})
+	refined_line = test_line.with_refined_pixel()
+
+Note that in this example and in the example below, ``calibration_spectrum`` is a
+`~specutils.Spectrum1D` spectrum to be used to refine the line locations (for example
+a lamp spectrum).
+
+1D Wavelength Calibration
+-------------------------
+
+Of course, refining the location of a line is only really useful when done for multiple
+lines to be fit as part of the dispersion model. This can be accomplished with the
+`~specreduce.wavelength_calibration.WavelengthCalibration1D` class as follows below.
+Note that the lines can be input as simple (wavelength, pixel) pairs rather than as
+`~specreduce.wavelength_calibration.CalibrationLine` objects - in this case the
+``default_refinement_method`` and ``default_refinement_kwargs`` will be applied to
+each input line::
+
+	import astropy.units as u
+	from specreduce import WavelengthCalibration1D
+	test_cal = WavelengthCalibration1D(calibration_spectrum,
+									   [(5000*u.AA, 0), (5100*u.AA, 10),
+									    (5198*u.AA, 20), (5305*u.AA, 30)],
+                                       default_refinement_method="gaussian")
+    calibrated_spectrum = test_cal.apply_to_spectrum(science_spectrum)
+
+The example above uses the default model (`~astropy.modeling.functional_models.Linear1D`)
+to fit the input spectral lines, and then applies the calculated WCS solution to a second
+spectrum (``science_spectrum``). Any other ``astropy`` model can be provided as the
+input ``model`` parameter to the `~specreduce.wavelength_calibration.WavelengthCalibration1D`.
+In the above example, the model fit and WCS construction is all done as part of the
+``apply_to_spectrum()`` call, but you could also access the `~gwcs.wcs.WCS` object itself
+by calling::
+
+	test_cal.wcs
+
+The calculated WCS is a cached property that will be cleared if the ``lines``, ``model``,
+or ``input_spectrum`` properties are updated, since these will alter the calculated dispersion
+fit.
--- a/docs/wavelength_calibration.rst
+++ b/docs/wavelength_calibration.rst
@@ -7,53 +7,25 @@ Wavelength calibration is currently supported for 1D spectra. Given a list of sp
 lines with known wavelengths and estimated pixel positions on an input calibration
 spectrum, you can currently use ``specreduce`` to:
 
-#. Refine the pixel position estimates of the lines based on features in the input
-   calibration spectrum.
 #. Fit an ``astropy`` model to the wavelength/pixel pairs to generate a spectral WCS
    solution for the dispersion.
 #. Apply the generated spectral WCS to other `~specutils.Spectrum1D` objects.
 
-Calibration Lines
------------------
-
-``specreduce`` provides a `~specreduce.wavelength_calibration.CalibrationLine` class for
-encoding the information about each spectral line to be used in the dispersion model
-solution fitting. The minimum information required is the wavelength and estimated pixel
-location on the calibration spectrum of each line. By default, no refinement of the pixel
-position is done; to enable this feature you must provide a refinement method and any
-associated keywords (currently ``range`` defines the number of pixels on either side
-of the estimated location to use for all available refinement options). For example,
-to define a line and use ``specreduce`` to update the pixel location the the centroid
-of a gaussian fit to the line, you would do the following::
-
-	import astropy.units as u
-	from specreduce import CalibrationLine
-	test_line = CalibrationLine(calibration_spectrum, 6562.81*u.AA, 500,
-							   refinement_method="gaussian",
-							   refinement_kwargs={"range": 10})
-	refined_line = test_line.with_refined_pixel()
-
-Note that in this example and in the example below, ``calibration_spectrum`` is a
-`~specutils.Spectrum1D` spectrum to be used to refine the line locations (for example
-a lamp spectrum).
-
 1D Wavelength Calibration
 -------------------------
 
-Of course, refining the location of a line is only really useful when done for multiple
-lines to be fit as part of the dispersion model. This can be accomplished with the
-`~specreduce.wavelength_calibration.WavelengthCalibration1D` class as follows below.
-Note that the lines can be input as simple (wavelength, pixel) pairs rather than as
-`~specreduce.wavelength_calibration.CalibrationLine` objects - in this case the
-``default_refinement_method`` and ``default_refinement_kwargs`` will be applied to
-each input line::
+The `~specreduce.wavelength_calibration.WavelengthCalibration1D` class can be used
+to fit a dispersion model to a list of line positions and wavelengths. Future development
+will implement catalogs of known lamp spectra for use in matching observed lines. In the
+example below, the line positions (``line_list``) have already been extracted from
+``lamp_spectrum``::
 
-	import astropy.units as u
-	from specreduce import WavelengthCalibration1D
-	test_cal = WavelengthCalibration1D(calibration_spectrum,
-									   [(5000*u.AA, 0), (5100*u.AA, 10),
-									    (5198*u.AA, 20), (5305*u.AA, 30)],
-                                       default_refinement_method="gaussian")
+    import astropy.units as u
+	 from specreduce import WavelengthCalibration1D
+	 line_list = [10, 22, 31, 43]
+	 wavelengths = [5340, 5410, 5476, 5543]*u.AA
+	 test_cal = WavelengthCalibration1D(lamp_spectrum, line_list,
+												  line_wavelengths=wavelengths)
     calibrated_spectrum = test_cal.apply_to_spectrum(science_spectrum)
 
 The example above uses the default model (`~astropy.modeling.functional_models.Linear1D`)
@@ -64,8 +36,17 @@ In the above example, the model fit and WCS construction is all done as part of 
 ``apply_to_spectrum()`` call, but you could also access the `~gwcs.wcs.WCS` object itself
 by calling::
 
-	test_cal.wcs
+    test_cal.wcs
 
-The calculated WCS is a cached property that will be cleared if the ``lines``, ``model``,
+The calculated WCS is a cached property that will be cleared if the ``line_list``, ``model``,
 or ``input_spectrum`` properties are updated, since these will alter the calculated dispersion
 fit.
+
+You can also provide the input pixel locations and wavelengths of the lines as an
+`~astropy.table.QTable`, with columns ``pixel_center`` and ``wavelength``::
+
+    from astropy.table import QTable
+    pixels = [10, 20, 30, 40]*u.pix
+    wavelength = [5340, 5410, 5476, 5543]*u.AA
+    line_list = QTable([pixels, wavelength], names=["pixel_center", "wavelength"])
+    test_cal = WavelengthCalibration1D(lamp_spectrum, line_list)
--- a/specreduce/conftest.py
+++ b/specreduce/conftest.py
@@ -6,6 +6,10 @@
 import os
 
 from astropy.version import version as astropy_version
+import astropy.units as u
+import numpy as np
+import pytest
+from specutils import Spectrum1D
 
 # For Astropy 3.0 and later, we can use the standalone pytest plugin
 if astropy_version < '3.0':
@@ -19,6 +23,33 @@ else:
     except ImportError:
         ASTROPY_HEADER = False
 
+@pytest.fixture
+def spec1d():
+    np.random.seed(7)
+    flux = np.random.random(50)*u.Jy
+    sa = np.arange(0, 50)*u.pix
+    spec = Spectrum1D(flux, spectral_axis=sa)
+    return spec
+
+@pytest.fixture
+def spec1d_with_emission_line():
+    np.random.seed(7)
+    sa = np.arange(0, 200)*u.pix
+    flux = (np.random.randn(200) +
+             10*np.exp(-0.01*((sa.value-130)**2)) +
+             sa.value/100) * u.Jy
+    spec = Spectrum1D(flux, spectral_axis=sa)
+    return spec
+
+@pytest.fixture
+def spec1d_with_absorption_line():
+    np.random.seed(7)
+    sa = np.arange(0, 200)*u.pix
+    flux = (np.random.randn(200) -
+             10*np.exp(-0.01*((sa.value-130)**2)) +
+             sa.value/100) * u.Jy
+    spec = Spectrum1D(flux, spectral_axis=sa)
+    return spec
 
 def pytest_configure(config):
 
--- a/specreduce/conftest.py
+++ b/specreduce/conftest.py
@@ -23,6 +23,7 @@ else:
     except ImportError:
         ASTROPY_HEADER = False
 
+
 @pytest.fixture
 def spec1d():
     np.random.seed(7)
@@ -31,25 +32,28 @@ def spec1d():
     spec = Spectrum1D(flux, spectral_axis=sa)
     return spec
 
+
 @pytest.fixture
 def spec1d_with_emission_line():
     np.random.seed(7)
     sa = np.arange(0, 200)*u.pix
     flux = (np.random.randn(200) +
-             10*np.exp(-0.01*((sa.value-130)**2)) +
-             sa.value/100) * u.Jy
+            10*np.exp(-0.01*((sa.value-130)**2)) +
+            sa.value/100) * u.Jy
     spec = Spectrum1D(flux, spectral_axis=sa)
     return spec
+
 
 @pytest.fixture
 def spec1d_with_absorption_line():
     np.random.seed(7)
     sa = np.arange(0, 200)*u.pix
     flux = (np.random.randn(200) -
-             10*np.exp(-0.01*((sa.value-130)**2)) +
-             sa.value/100) * u.Jy
+            10*np.exp(-0.01*((sa.value-130)**2)) +
+            sa.value/100) * u.Jy
     spec = Spectrum1D(flux, spectral_axis=sa)
     return spec
+
 
 def pytest_configure(config):
 
--- a/specreduce/wavelength_calibration.py
+++ b/specreduce/wavelength_calibration.py
@@ -9,6 +9,9 @@ from specutils import SpectralRegion, Spectrum1D
 from specutils.fitting import fit_lines
 
 
+__all__ = ['CalibrationLine', 'WavelengthCalibration1D']
+
+
 class CalibrationLine():
 
     def __init__(self, input_spectrum, wavelength, pixel, refinement_method=None,
--- a/specreduce/wavelength_calibration.py
+++ b/specreduce/wavelength_calibration.py
@@ -1,5 +1,7 @@
 from astropy.modeling.models import Linear1D, Gaussian1D
 import astropy.units as u
+from functools import cached_property
+import numpy as np
 from specutils import SpectralRegion
 from specutils.fitting import fit_lines
 
@@ -10,7 +12,7 @@ Starting from Kyle's proposed pseudocode from https://github.com/astropy/specred
 
 class CalibrationLine():
 
-    def __init__(input_spectrum, wavelength, pixel, refinement_method=None,
+    def __init__(self, input_spectrum, wavelength, pixel, refinement_method=None,
                  refinement_kwargs={}):
         """
         input_spectrum: `~specutils.Spectrum1D`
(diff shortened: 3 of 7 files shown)


specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+            raise ValueError("You must define 'direction' in refinement_kwargs to use "
+                             "gradient refinement")
+
+    @classmethod
+    def by_name(cls, input_spectrum, line_name, pixel,
+                refinement_method=None, refinement_kwargs={}):
+        # this will do wavelength lookup and passes that wavelength to __init__ allowing the user
+        # to call CalibrationLine.by_name(spectrum, 'H_alpha', pixel=40)
+        # We could also have a type-check for ``wavelength`` above, this just feels more verbose
+        return cls(...)
```
kecnry, line 55: should we just remove this for now?

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+        self.refinement_method = refinement_method
+        self.refinement_kwargs = refinement_kwargs
+
+        if self.refinement_method in ("gaussian", "min", "max"):
+            if 'range' not in self.refinement_kwargs:
+                raise ValueError(f"You must define 'range' in refinement_kwargs to use "
+                                 f"{self.refinement_method} refinement.")
+        elif self.refinement_method == "gradient" and 'direction' not in self.refinement_kwargs:
+            raise ValueError("You must define 'direction' in refinement_kwargs to use "
+                             "gradient refinement")
```
kecnry, line 47: do we want defaults for these options or is it too hard to make reasonable defaults that can be generalized?

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+        input_spectrum = self.input_spectrum if input_spectrum is None else input_spectrum
+        refined_pixel = self._do_refinement(input_spectrum)
+        if return_object:
+            return CalibrationLine(input_spectrum, self.wavelength, refined_pixel,
+                                   refinement_method=self.refinement_method,
+                                   refinement_kwargs=self.refinement_kwargs)
+
+        return refined_pixel
+
+    @cached_property
```
kecnry, line 100: should probably be coupled with custom setters for the input spectrum and options to clear the cache

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+    def refined_pixel(self):
+        # returns the refined pixel for self.input_spectrum
+        return self.refine()
+
+    @property
```
kecnry, line 105: could also be cached (see comment above)

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+        return f"CalibrationLine: ({self.wavelength}, {self.pixel})"
+
+    def __repr__(self):
+        return f"CalibrationLine({self.wavelength}, {self.pixel})"
+
+
+class WavelengthCalibration1D():
+
+    def __init__(self, input_spectrum, lines, model=Linear1D(), spectral_unit=u.Angstrom,
+                 fitter=LMLSQFitter(calc_uncertainties=True),
```
kecnry, line 120: should we try to avoid the astropy warning by using a different default fitter, or maybe by detecting whether `model` is linear and choosing appropriately?

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+                else:
+                    self.lines.append(CalibrationLine(self.input_spectrum, line[0], line[1],
+                                      self.default_refinement_method,
+                                      self.default_refinement_kwargs))
+
+    @classmethod
+    def autoidentify(cls, input_spectrum, line_list, model=Linear1D):
+        # line_list could be a string ("common stellar") or an object
+        # this builds a list of CalibrationLine objects and passes to __init__
+        return cls(...)
```
kecnry, line 169: remove from code for now?

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+    @property
+    def refined_lines(self):
+        return [line.with_refined_pixel for line in self.lines]
+
+    @property
+    def refined_pixels(self):
+        # useful for plotting over spectrum to change input lines/refinement options
+        return [(line.wavelength, line.refined_pixel) for line in self.lines]
+
+    @cached_property
```
kecnry, line 180: cache needs to be cleared when appropriate

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,248 @@
...
+        pixel_frame = cf.CoordinateFrame(1, "SPECTRAL", [0,], axes_names=["x",], unit=[u.pix,])
+        spectral_frame = cf.SpectralFrame(axes_names=["wavelength",], unit=[self.spectral_unit,])
+
+        pipeline = [(pixel_frame, self.model), (spectral_frame, None)]
+
+        wcsobj = wcs.WCS(pipeline)
+
+        return wcsobj
+
+    def apply_to_spectrum(self, spectrum=None):
```
kecnry, line 199: Do we want `__call__` instead, to just call this, to do something else, or to not exist?  I think we will eventually want calling to do _something_ since that pattern is supported through most/all of the other specreduce operations.

kecnry (approved):

specreduce/wavelength_calibration.py
rosteen: I remembered @eteq was somewhat against having a `__call__` at all, so I had left it out for now. If it's more consistent with the way the rest of the package works, then I guess we can think about what makes the most sense. Having it call this method on the default (input) spectrum seems like an ok behavior to me, but I don't feel strongly about it.

specreduce/wavelength_calibration.py
rosteen: That can be accomplished by deleting the cached properties from the class `__dict__`, right? I haven't use `cached_property` before

kecnry: exactly. In jdaviz I wrote a [helper method](https://github.com/spacetelescope/jdaviz/blob/main/jdaviz/core/template_mixin.py#L242-L250) that might also be useful in this case

specreduce/wavelength_calibration.py
rosteen: Would you prefer to do that, or raise a NotImplementedError?

rosteen: Fair, I'll just remove it.

specreduce/wavelength_calibration.py
rosteen: Good call, I updated to check whether the model is linear or not and use a different fitter accordingly

specreduce/wavelength_calibration.py
rosteen: Alright, I implemented what I think you had in mind, take a look.

specreduce/wavelength_calibration.py
rosteen: I would be lean toward making gaussian the default, personally.

rosteen: The issue is that the `range` parameter depends heavily on the properties of your spectrum/array, I think. Hence why I left the default as no refinement.

kecnry: Sorry, I meant automatic defaults for `refinement_kwargs`, definitely `refinement_method`-dependent, and perhaps even estimated from the `input_spectrum` (not sure how complicated that would get - definitely could be a follow-up effort... I just feel like users are constantly going to be seeing this error message).

specreduce/wavelength_calibration.py
kecnry: that looks good to me, but I think there are other cases that need to clear too (`wcs` cache needs to clear for a change in pretty much any of the original inputs, I think).  I wonder if there's a way to generalize away this logic so we don't need to have all this boilerplate code 🤔 (definitely can generalize later - it would probably take a custom `__setattr__` which might be a little hacky-feeling)

specreduce/wavelength_calibration.py
kecnry: Ok, I think we'll want `__call__` to do _something_ eventually, but I'm fine deferring that until the rest of the API begins to converge in follow-up efforts.

specreduce/wavelength_calibration.py
rosteen: Also now clears the WCS cache on changing the `lines` or `model` attributes. Do you think that suffices for now? I imagine that at some point people would just make a new `WavelengthCalibration1D` rather than updating an existing one.

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,303 @@
...
+        return [(line.wavelength, line.refined_pixel) for line in self.lines]
+
+    @cached_property
+    def wcs(self):
+        # computes and returns WCS after fitting self.model to self.refined_pixels
+        x = [line[1] for line in self.refined_pixels] * u.pix
+        y = [line[0].value for line in self.refined_pixels] * self.spectral_unit
+
+        # Fit the model
+        self._model = self.fitter(self._model, x, y)
```
kecnry, line 242: does this prevent successive fitting calls?

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,303 @@
...
+        self.spectral_unit = spectral_unit
+        self.default_refinement_method = default_refinement_method
+        self.default_refinement_kwargs = default_refinement_kwargs
+        self._cached_properties = ['wcs',]
+
+        if fitter is None:
+            if self.model.linear:
+                self.fitter = LinearLSQFitter(calc_uncertainties=True)
+            else:
+                self.fitter = LMLSQFitter(calc_uncertainties=True)
```
kecnry, line 164: what if instead `self._fitter` was stored as `None` and then a `fitter` property defaults based on the _current_ value of `self.model` (so that if `model` is changed after initialization, the default can change to match)?

rosteen: I like this idea, I'll implement momentarily.

specreduce/wavelength_calibration.py
rosteen: I don't think so...I guess successive calls would be starting from the previous fitted parameter values, but I don't think it prevents fitting. Do you think having a separate `_fitted_model` parameter would be better?

kecnry: I'm not sure 🤷, I think the way you have it - assuming the output can be passed as the input - is both quite handy but also maybe unexpected behavior unless we document carefully.

rosteen:
@kecnry I added docs - please look them over. I'm going to open an issue to revisit whether we want to store the fitted model in a separate property from the initial model.

tepickering (commented):
a few high-level comments:

- there definitely needs to be a concept of a `LineList`. for example, the `HeNeAr` lamps that are very commonly used for ground-based optical data provide several hundred possible calibration lines. different calibration lamps are often combined so there needs to be a way to manage and combine lists based on configuration. handling such sets of calibration lines is a far more common use case than managing individual ones.
- i'm not sure why the `CalibrationLine` class needs pixel position refinement. the locating and centroiding of potential calibration lines in pixel-space should happen in a separate, previous step. 
- iterative fitting is something we'll likely want to implement. basically, start with `min_order` and increase `order` until either the fit RMS stops improving or `max_order` is reached. it's probably something that can be added later, but we should be cognizant of the future need.

rosteen:
Thanks for the thoughts, Tim. Your points 1 and 3 are planned for future development, we're just trying to be iterative about this and not have one massive PR that takes 6 months to finish. Your second comment is the only one that might be an actual sticking point, I think, but I don't think that the current class precludes the workflow it sounds like your imagining. If the pixel positions of the lines have already been refined, you can simply leave the refinement args in the `CalibrationLine` class to their defaults, which don't do any refinement. As you mentioned with the `LineList` point, the plan is to add more automated options later, but this initial work enables the workflow of manually identifying some lines in the cal spectrum and using that starting point to do the fit (with or without refinement). I see the workflow you seem to want as compatible enhancements to this initial work, not as fundamentally incompatible. Do you disagree?

tepickering:
my main concern, especially with point 2, is that sub-optimal architecture that gets baked in early causes significant issues down the road. wavelength calibration fundamentally requires managing lists of lines because you need multiple lines to formulate a wavelength solution. as implemented here, the `CalibrationLine` class is mixing pixel and wavelength space for each individual line. it might make sense for the simple, manual examples here, but i'm not seeing how it will be at all beneficial when dealing with large lists of lines and no prior knowledge of pixel->wavelength mapping. 

in my head and how most other implementations work, you first create an extracted line list from an extracted spectrum. the peak-finding and accurate centroiding of the lines is done in the creation of that. this can fundamentally just be a list of pixel positions, but could optionally include line fluxes as well. the fluxes could be used to set thresholds for what lines to use or can be used by matching algorithms. 

then you have a list of wavelengths of potential calibration lines. usually these are archived catalogs, but it could be a manual list. again, fundamentally just a list of wavelengths, but could include other optional info like relative strengths or whether the wavelengths are vacuum or air. 

i should note that `pypeit` and `DRAGONS` use `astropy.table` internally for handling these kinds of lists. we may want to do likewise because i can definitely foresee potential needs for functionality like `vstack`. 

where a `CalibrationLine` class might come into play is where lines are matched between the pixel list and wavelength list. however, something list-like would make more sense since you're passing a list of wavelength/pixel pairs to the fitting routine. iterating through a bunch of `CalibrationLine` instances to get that information would be a _lot_ slower than just grabbing a couple of lists or numpy arrays from a `Table` or similar.

this matching is a process and there are many ways to do it. this initial example of manual matching is fine and there will always be a use case for that. my contribution will be porting over automated and template-based matching. the output from any of the methods should be the same format of wavelength/pixel pairs. so there should be some kind of class hierarchy of `LineMatcher` or something like that to cover these. maybe not now, but definitely once i add my stuff. 

i'll see if i can put these words into a clearer flowchart...

rosteen:
Alright, as requested offline here's the workflow I had in mind, for posterity:

- User has spectrum open in some GUI tool (e.g., Specviz).
- In GUI, select lines in the spectrum to be used for calibration by selecting subsets around the lines (or some other UI, perhaps by defining the threshold).
- User inputs the wavelengths for the lines, knowing what they are from some sort of prior knowledge.
- Specreduce gets called through the UI to centroid the pixel positions of the lines within the defined subsets and run the fit based on pixel locations and wavelengths, assigning the derived WCS so that the x-axis in the GUI can be plotted as wavelength instead of pixels.

Based on @tepickering's  comments, I think I just need to shuffle things around a bit - we can take the list of refined pixel positions and list of (potential) lines as input to the `WavelengthCalibration1D` class, which will then apply the matching algorithm, and translate the selected lines to a list of refined locations on the GUI application side. In that case, the `CalibrationLine` class is unneeded, since we'll be storing all that information in tables in the `WavelengthCalibration1D` class. 

The thing I'm still torn in is whether we should keep the option to refine the pixel positions here. If the user does their refining/centroiding elsewhere that's fine, they don't have to use it, but I do think it would be convenient to have. Maybe it makes sense to have a convenience method for `specutils` line finding as well, since line finding/centroiding do seem like a logical grouping. Probably makes sense to be able to do both or neither rather than splitting them. 

@cpacifici @kecnry @tepickering @eteq thoughts? And should I reopen this against a feature branch so that these decisions are slightly less impactful in the short term and we have more time to iron these sorts of things out?

camipacifici:
I do not have a good sense of how the code would change, sorry, but as long as the science case you wrote (which I share as my usual science case for wavelength calibration) is supported, I am happy.
@tepickering @eteq?

tepickering:
my main point is that peak-finding/line-centroiding is a _very_ separate process from wavelength calibration. the code organization should reflect this or there will be future pain. the flow should be:

- load spectrum into gui
- click something like 'find lines' to locate and centroid features in the spectrum. adjust parameters and iterate as needed. parameters that may need adjustment include centroiding method (moments vs gaussian fit), line fwhm and whether that's fixed or fit, continuum subtraction or not, thresholding, saturation limit. when you get to dealing with messy, real-world data, those details can become important. we don't need to deal with them all now, but need to have a framework for doing so going forward. 
- list of pixel positions from previous step gets displayed on gui, user clicks on id'ed lines with known wavelengths, and provides the wavelengths. 
- list of matched pixel/wavelength passed to `specreduce`, wavelength calibration calculated, WCS passed back to gui.
- with wavelength scale now available in gui, more lines can be id'ed by hand or via automated matching and the process iterated to refine the wavelength calibration.

peak-finding and line-centroiding themselves can be separated and usually are. having them separate would support the proposed flow of manual peak-finding via a gui with refinement by `specreduce`. however, lots of implementations of the automated approach exist and could be relatively easily adopted. an obvious candidate would be https://specutils.readthedocs.io/en/stable/api/specutils.fitting.find_lines_threshold.html#specutils.fitting.find_lines_threshold. 

i should note that the wheel being reinvented here is described at https://pypeit.readthedocs.io/en/release/calibrations/wave_calib.html#by-hand-approach. there is interest on that end to replace their Qt-based guis with web/browser-based ones. collaboration towards that would be ultimately beneficial to all so we should keep that in mind.

rosteen:
@tepickering I just pushed a bunch of changes to move in the direction you described, mind taking a look? I updated the tests as well but the docs still need to be updated to reflect these changes.

tepickering:
awesome! i'll take a quick look tonight and a deeper dive tomorrow...

tepickering:
the `readthedocs` build error should get fixed once this is rebased to include #165 after it's merged. i am deprecating `specreduce_data` as a python package and making into simply a data repository.

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,228 @@
...
+            named "pixel_center" and optionally a "wavelength" column with known line
+            wavelengths populated.
+        line_wavelengths: `~astropy.units.Quantity`, `~astropy.table.QTable`, optional
+            `astropy.units.Quantity` array of line wavelength values corresponding to the
+            line_pixels. Does not have to be in the same order (the lists will be sorted)
+            but does currently need to be the same length as line_list. Can also be input
+            as an `~astropy.table.QTable` with (minimally) a "wavelength" column.
+        catalog: list, str, optional
+            The name of a catalog of line wavelengths to load and use in automated and
+            template-matching line matching.
```
kecnry, line 50: could this be moved to a class method?  So `WavelengthCalibration1D.from_line_list(line_list)` calls `WavelengthCalibration1D(...)`?

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,228 @@
...
+            A one-dimensional Spectrum1D calibration spectrum from an arc lamp or similar.
+        line_list: list, array, `~astropy.table.QTable`
+            List or array of line pixel locations to anchor the wavelength solution fit.
+            Will be converted to an astropy table internally if a list or array was input.
+            Can also be input as an `~astropy.table.QTable` table with (minimally) a column
+            named "pixel_center" and optionally a "wavelength" column with known line
+            wavelengths populated.
+        line_wavelengths: `~astropy.units.Quantity`, `~astropy.table.QTable`, optional
+            `astropy.units.Quantity` array of line wavelength values corresponding to the
+            line_pixels. Does not have to be in the same order (the lists will be sorted)
```
kecnry, line 45:
```suggestion
            line pixels defined in ``line_list``. Does not have to be in the same order (the lists will be sorted)
```

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,228 @@
...
+                                 "a 'wavelength' column in line_list.")
+            if len(line_wavelengths) != len(line_list):
+                raise ValueError("If line_wavelengths is specified, it must have the same "
+                                 "length as line_pixels")
+            if not isinstance(line_wavelengths, (u.Quantity, QTable)):
+                raise ValueError("line_wavelengths must be specified as an astropy.units.Quantity"
+                                 "array or as an astropy.table.QTable")
+            if isinstance(line_wavelengths, u.Quantity):
+                line_wavelengths.value.sort()
+                self._line_list["wavelength"] = line_wavelengths
```
kecnry, line 93: does `line_wavelengths` support frequencies?  Would the sorting need to be flipped in this case?  What if they're mixed units?

specreduce/conftest.py
```diff
@@ -20,6 +24,37 @@
...
+def spec1d_with_absorption_line():
+    np.random.seed(7)
+    sa = np.arange(0, 200)*u.pix
+    flux = (np.random.randn(200) -
+            10*np.exp(-0.01*((sa.value-130)**2)) +
+            sa.value/100) * u.Jy
+    spec = Spectrum1D(flux, spectral_axis=sa)
+    return spec
+
+
```
kecnry, line 57: would this benefit from using the synthetic spectra from #165?

kecnry (commented):
Just a few comments/questions from a quick skim of the code.  Are there any examples for different workflows that can be used for testing or should we just refer to the tests for now?

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,228 @@
...
+                self._catalog = catalog
+            else:
+                self._catalog = [catalog]
+            for cat in catalog:
+                if isinstance(cat, str):
+                    if cat not in self._available_catalogs:
+                        raise ValueError(f"Line list '{cat}' is not an available catalog.")
+
+            # Get the potential lines from any specified catalogs to use in matching
+            self._potential_wavelengths = concatenate_catalogs(self._catalog)
```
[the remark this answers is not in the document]
rosteen: Trying to input anything for `catalog` would get caught by the `ValueError` above since there aren't any available yet, but I agree that it would be better as a `NotImplementedError` for now.

specreduce/wavelength_calibration.py
rosteen: I'm not sure how common it is to use frequencies for this (my assumption was "never" but I guess _someone_ might want to), but we could support it. I can also add a check to make sure they're all in the same unit.

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,228 @@
...
+            `astropy.units.Quantity` array of line wavelength values corresponding to the
+            line_pixels. Does not have to be in the same order (the lists will be sorted)
+            but does currently need to be the same length as line_list. Can also be input
+            as an `~astropy.table.QTable` with (minimally) a "wavelength" column.
+        catalog: list, str, optional
+            The name of a catalog of line wavelengths to load and use in automated and
+            template-matching line matching.
+        model: `~astropy.modeling.Model`
+            The model to fit for the wavelength solution. Defaults to a linear model.
+        """
```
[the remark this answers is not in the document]
rosteen: I actually meant to remove `spectral_unit` since we taking `Quantity` or `QTable` for the spectral values now.

specreduce/conftest.py
rosteen: Yes, but I think using that to expand/improve the tests can be another PR once both of these are merged.

specreduce/wavelength_calibration.py
rosteen: Actually, since this requires `line_wavelengths` to be a `Quantity` or `QTable` (rather than allowing a list of `Quantity`), it's already guaranteed that all elements have the same unit.

rosteen: @kecnry I implemented reverse-sorting if frequencies are provided, although that does make the variable name `line_wavelengths` a bit of a misnomer.

rosteen:
@tepickering Think you'll get a chance to take a deeper look by Monday afternoon or so? I'd like to close this out in line with our Sprint end.

specreduce/wavelength_calibration.py
tepickering: my first take is that since this package is geared towards 2D optical/IR data obtained with CCD-like devices, i would stick with wavelengths internally and not worry about frequencies at all. at least for simplicity at this stage of development. if users have dispersed data with units of frequency, there are existing utilities to transform it to wavelengths. i can imagine possible use cases for 2D spatial vs frequency data in radio astronomy, but i've never seen it. radio data is either 1D (single-dish) or 3D (scanned single-dish or interferometry). eV or wavenumber might be more likely candidates for non-wavelength inputs, but they can also easily be transformed to wavelength.

tepickering:
i will take a deeper dive later today, if not over the weekend. it was a little hard to follow the changes on github due to missing test coverage so i'll pull in a local copy.

docs/wavelength_calibration.rst
```diff
@@ -0,0 +1,52 @@
...
+	 from specreduce import WavelengthCalibration1D
+	 line_list = [10, 22, 31, 43]
+	 wavelengths = [5340, 5410, 5476, 5543]*u.AA
+	 test_cal = WavelengthCalibration1D(lamp_spectrum, line_list,
+												  line_wavelengths=wavelengths)
+    calibrated_spectrum = test_cal.apply_to_spectrum(science_spectrum)
+
+The example above uses the default model (`~astropy.modeling.functional_models.Linear1D`)
+to fit the input spectral lines, and then applies the calculated WCS solution to a second
+spectrum (``science_spectrum``). Any other ``astropy`` model can be provided as the
```
tepickering, line 33: maybe specify "any 1D model"

docs/wavelength_calibration.rst
```diff
@@ -0,0 +1,52 @@
...
+or ``input_spectrum`` properties are updated, since these will alter the calculated dispersion
+fit.
+
+You can also provide the input pixel locations and wavelengths of the lines as an
+`~astropy.table.QTable`, with columns ``pixel_center`` and ``wavelength``::
+
+    from astropy.table import QTable
+    pixels = [10, 20, 30, 40]*u.pix
+    wavelength = [5340, 5410, 5476, 5543]*u.AA
+    line_list = QTable([pixels, wavelength], names=["pixel_center", "wavelength"])
```
tepickering, line 51: nitpick: i'd call this `matched_line_list` so as not to be confused with `linelist` or `line_list` which is usually understood to mean something else (i.e. list of wavelengths of lines due to specific ion/molecule or mixture of of them).

tepickering (changes requested):
this is a much-improved implementation that gives us something to build on. i left some minor comments about some wording that should get sorted out. there'll be some meshing needed between this and what i've done in #165. that can happen in a future PR, though. the most substantive thing i'd like to see addressed before merging to cover the remaining missing lines with tests. there's not too many...

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,200 @@
...
+          and write this function to determine the list of available catalog names.
+    """
+    return []
+
+
+def concatenate_catalogs():
+    """
+    ToDo: Code logic to combine the lines from multiple catalogs if needed
+    """
+    pass
```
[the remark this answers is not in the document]
rosteen: Great, we'll definitely need a PR after these are both merged to combine their powers 😃

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+class WavelengthCalibration1D():
+
+    def __init__(self, input_spectrum, line_list, line_wavelengths=None, catalog=None,
+                 model=Linear1D(), fitter=None):
+        """
+        input_spectrum: `~specutils.Spectrum1D`
+            A one-dimensional Spectrum1D calibration spectrum from an arc lamp or similar.
+        line_list: list, array, `~astropy.table.QTable`
```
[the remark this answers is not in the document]
rosteen: My thought here was that the user can provide the whole table, including both pixels and wavelengths, in which case it doesn't make sense to call it `'line_pixels`. Maybe I'll change this slightly to have separate `line_list` and `line_pixels` arguments, where the former can be the whole table, and the latter is intended for just pixel values to then put internally into the table.

rosteen:
Coverage is up to 90%, the rest is really dependent on integrating with #165 so I think it can wait. I think all I have left to address @tepickering's comments is to split out `matched_line_list` and `line_pixels` into separate input arguments - I think the way forward is to make them both optional but require at least one.

tepickering (approved):

tepickering:
i'm happy with the latest round of changes. as previously mentioned, further work to mesh functionality and vocabulary with #165 is best left to a subsequent PR.

rosteen:
Excellent! Tests passed again after my update splitting out those two input arguments, let's get this in 🎉

tepickering:
lgtm! if @kecnry concurs, feel free to merge. then i'll do the same for #165.

specreduce/wavelength_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+        model: `~astropy.modeling.Model`
+            The model to fit for the wavelength solution. Defaults to a linear model.
+        fitter: `~astropy.modeling.fitting.Fitter`, optional
+            The fitter to use in optimizing the model fit. Defaults to
+            `~astropy.modeling.fitting.LinearLSQFitter` if the model to fit is linear
+            or `~astropy.modeling.fitting.LMLSQFitter` if the model to fit is non-linear.
+
+        Note that either ``matched_line_list`` or ``line_pixels`` must be specified,
+        and if ``matched_line_list`` is not input, at least one of ``line_wavelengths``
+        or ``catalog`` must be specified.
```
kecnry, line 64: this feels to me like it would be better as multiple classmethod, but we can revisit this as we iterate further on the API and various workflows as I recognize that might be a matter of taste:

* `WavelengthCalibration1D(input_spectrum, matched_line_list, model, fitter)` (`__init__`) 
* `WavelengthCalibration1D.from_lists(input_spectrum, line_pixels, line_wavelengths, model, fitter)`
* `WavelengthCalibration1D.from_catalog(input_spectrum, catalog, model, fitter)` (and perhaps `line_pixels`?)

This also would allow getting rid of a lot of code that just checks for validity of the combinations of all of these technically-optional but-sorta-required kwargs.

kecnry (commented):
I have some questions/concerns about the big-picture plan for this as catalogs and the 2d cases get implemented, but so long as the plan is to iterate further in the future, won't object to merging this as-is.  I'd love to see the actual pseudocode examples updated for various science workflows though - just to see where this is heading in advance.  @tepickering, how do you imagine the catalog interface would work with this?  There is no refinement implemented here anymore, so how does the 2d case work?

I know we haven't committed to backwards-compatibility yet, but would it make sense to merge this into a development branch instead of main until we're ready for it to be (more) public facing (@tepickering)?

specreduce/wavelength_calibration.py
tepickering: yeah, this is still kinda messy. i think we should actually keep `WavelengthCalibration1D` as simple as possible: just `WavelengthCalibration1D(line_pixels, line_wavelengths, model, fitter)` where `line_pixels` and `line_wavelengths` are list-like of the same length. there are previous steps required to find/centroid lines in a calibration spectrum (`specutils` provides tools for this) and then match some of those lines to known wavelengths to construct `line_pixels` and `line_wavelengths` (manually or via to-be-implemented automated techniques). i don't think we should wrap all of these steps into one big class. the interfacing with things like catalogs of known lines like i implement in #165 will happen at the line-matching stage. 

anyway, it will be a lot easier to iterate on these ideas using realistic end-to-end examples once this work is merged with #165. the work may not be complete after these are merged, but nothing is broken so i don't think we need a separate `develop` branch`.

docs/wavelength_calibration.rst
```diff
@@ -0,0 +1,53 @@
...
+to fit a dispersion model to a list of line positions and wavelengths. Future development
+will implement catalogs of known lamp spectra for use in matching observed lines. In the
+example below, the line positions (``pixel_centers``) have already been extracted from
+``lamp_spectrum``::
+
+    import astropy.units as u
+	 from specreduce import WavelengthCalibration1D
+	 pixel_centers = [10, 22, 31, 43]
+	 wavelengths = [5340, 5410, 5476, 5543]*u.AA
+	 test_cal = WavelengthCalibration1D(lamp_spectrum, line_pixels=pixel_centers,
```
tepickering, line 27: note that `lamp_spectrum` is not defined in this example.

specreduce/wavelength_calibration.py
tepickering: a workflow we will need to support is where we iterate the fitting process. one would start with a small number of matched lines and a simple model, use the model to match more lines, re-fit, tweak model, match more lines, re-fit/tweak/match, and repeat until a convergence criteria is met. my view is that this class/function should be within that loop, but not contain that loop. i don't see where a ``__call__`` fits into that scheme and i think a `apply_to_spectrum()` method is a clearer way of applying the results to a spectrum.

tepickering (approved):
i've left several more substantive comments. i still think `WavelengthCalibration1D` is a bit too high-level, though a big step in the right direction. it works well enough to merge now and future work will be a lot easier with #165 merged in as well. being able to generate realistic data to prototype realistic workflows will be a significant benefit.

PR #163, opened 2023-02-01
Modify masking and NaN handling in `HorneExtract`

### LINKED ISSUE #158: NaNs outside background window still propagate into 1D spectra

ojustino:
This pull request is part 2 of a response that would close #158. (See #159 for part 1.)

### Goal:
When an image column contains NaNs, `HorneExtract` either errors out or causes the illusion of an absorption feature because the column has at least a pixel missing compared to others. An ideal solution would be to re-normalize each column to account for missing pixels, but for now, the priority is to prevent errors and potentially leading users astray.

In response, I have modified `HorneExtract` to better account for NaNs and updated the use of masks to account for the different approach.

### Roadblock:
Unlike masked `numpy` arrays, it looks like `Spectrum1D` objects don't hide masked values or even indicate which indices are masked by default. (You have to specifically call the `mask` attribute to find the latter.)

Instead, initializing a `Spectrum1D` object from a masked array gives me zero flux at indices of the 1D extraction that should be masked. This leaves the opportunity open for a misinterpreted absorption feature. [I see a related issue](https://github.com/astropy/specutils/issues/585) in `specutils` that hasn't been resolved.

Returning NaNs at indices that should be masked seems bad for users who want to do arithmetic with the extracted spectrum. We see enough other benefits with `Spectrum1D` that changing convention by returning a masked array instead also doesn't seem good. **Are there any other ideas besides pushing for a change in `specutils`?**

_(@kecnry: Since plots also show zero flux at affected indices, this issue also affects Specviz2D. However, if we can't solve this issue, I think you could still work with the 1D extraction's `mask` attribute behind the scenes to hide values that should be masked.)_

### Other questions:
- **Should rows with one or more non-finite values be masked in `coadd`?** (In this PR, they are.) If so, such rows are not factored into the Gaussian model used to fit each column's flux.
- **Should the extracted flux of a column with one or more non-finite values be accessible by users or completely masked?** (In this PR, the intention was to mask it, though it's not currently implemented due to my findings with `Spectrum1D`.)

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -535,42 +535,50 @@ class HorneExtract(SpecreduceOperation):
 
         # use compound model to fit a kernel to each fully finite image column
         # NOTE: infers Gaussian1D source profile; needs generalization for others
-        col_mask = np.logical_or.reduce(or_mask, axis=crossdisp_axis, dtype=bool)
+        col_mask = np.logical_or.reduce(or_mask, axis=crossdisp_axis)
+        nonf_col = [np.nan] * img.shape[crossdisp_axis]
 
         kernel_vals = []
         norms = []
         for col_pix in range(img.shape[disp_axis]):
+            # for now, skip columns with any non-finite values
+            # NOTE: fit and other kernel operations should support masking again
+            # once a fix is in for renormalizing columns with non-finite values
+            if col_mask[col_pix]:
+                kernel_vals.append(nonf_col)
+                norms.append(np.nan)
+                continue
+
             # else, set compound model's mean to column's matching trace value
             fit_ext_kernel.mean_0 = mean_init_guess[col_pix]
 
             # NOTE: support for variable FWHMs forthcoming and would be here
 
-            # fit compound model to column and mask rows with non-finite values
-            fitted_col = np.ma.masked_array(fit_ext_kernel(xd_pixels),
-                                            mask=col_mask[col_pix])
+            # fit compound model to column
+            fitted_col = fit_ext_kernel(xd_pixels)
 
             # save result and normalization
             kernel_vals.append(fitted_col)
             norms.append(fit_ext_kernel.amplitude_0
                          * fit_ext_kernel.stddev_0 * np.sqrt(2*np.pi))
 
-        # transform fit-specific information while preserving mask
-        kernel_vals = np.ma.vstack(kernel_vals).T
+        # transform fit-specific information
+        kernel_vals = np.vstack(kernel_vals).T
         norms = np.array(norms)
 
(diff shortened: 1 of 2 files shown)


ojustino:
@kecnry and I now think that returning NaNs at the indices I was trying to mask earlier is the best temporary approach. It's consistent with #159, where any in-window NaNs propagate through to boxcar-extracted 1D spectra. The window is the entire image in Horne extraction, so I don't mask NaNs at all in Horne-extracted 1D spectra.

tepickering (approved):
handling NaN's is such a pain and there's no ideal way to do so. for the lack of a better way, i generally agree with the `specutils` approach of keeping the original data intact and having the mask separate. as such, we should follow along accordingly.

kecnry (approved):
I think this is a good solution (at least for now).  It behaves similarly to Boxcar extraction (except of course that has a defined window, so a single nan in a column doesn't _necessarily_ result in a nan in the output spectrum, whereas it does here).  In the future, we can look into renormalization options for either/both of these, but this is a clear improvement to the previous confusing behavior.  Thanks!

ojustino:
Thank you for the reviews.

If you eventually decide to return to masking, another option besides `np.ma.masked_array()` might be [Astropy's own Masked class](https://docs.astropy.org/en/latest/utils/masked/index.html#utils-masked-vs-numpy-maskedarray) (h/t @bmorris3). It's currently  experimental, but perhaps that means `specreduce` could have some influence on how it interacts with `Spectrum1D` objects to avoid the roadblock in the original post.

PR #166, opened 2023-03-15
Improve Horne docs

camipacifici:
Added to the docs examples for Horne extract, the caveat that it needs uncertainty and mask, and the note that it works best for sources with a Gaussian profile.

[Generated docs](https://specreduce--166.org.readthedocs.build/en/166/extraction_quickstart.html)


kecnry (approved):
This is a nice improvement, thanks!

I don't expect this to be controversial and it only affects docs, so I'll merge.

PR #173, opened 2023-04-27
Add option to fit and use self profile for HorneExtract

cshanahan1:
This PR adds the 'self_profile' option to HorneExtract. Closes JDAT-3222.

A summary of how this works (row/col language assuming a horizontal trace):
- The image is divided along the dispersion axis into equal (plus minus one to make integer sized bins) bins along the x axis.
- Columns with masked or NaN values are omitted (see comments below, proper masking should be addressed eventually)
- The median in each bin along the x axis is computed, producing a median profile in that bin
- These samples are interpolated between (using RectBivariateSpline, by default linearly)
- The Horne normalization process now happens for each column. The interpolator is evaluated at that wavelength column to produce the spatial profile used for weighting. A 1D spectrum object is returned.

I also expanded the docs and fixed a few things in there (i.e a mask is not strictly required for HorneExtract as implied before, but is optional and currently limited in its functionality) 

A few comments, concerns I have for reviewers:
1. This, like the default Gaussian option, does not handle masking properly. This should be addressed in a follow up PR but for now I kept the same limitation that if there are any masked values in a wavelength column, the 1D spectrum will not be returned for that wavelength.
2. Should we return the fit profile (a 2D interpolator object, or the interpolator evaluated on the image grid) as an attribute on the extract object? Right now I have it as a private attribute for testing purposes, but I'm not sure if this is something users would go looking for?
4. The Gaussian option allows you to provide 'bkg_model' to fit simultaneously with the Gaussian extraction kernel, creating a compound model. I did not implement this for the self_profile option because it was not straightforward since I'm no longer using models/fitters and couldn't create a compound model to fit. I also disagree with this option in general, as the documentation and existence of background steps in Specreduce implies that background should be removed beforehand, so I didn't want to spend too much time figuring this out in the event that others agree with my opinion on this. If not, I will figure out how to add this option in a follow up PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@
 New Features
 ^^^^^^^^^^^^
 
+- Added 'interpolated_profile' option for HorneExtract. If The ``interpolated_profile`` option
+is used, the image will be sampled in various wavelength bins (set by
+``n_bins_interpolated_profile``), averaged in those bins, and samples are then
+interpolated between (linear by default, interpolation degree can be set with
+the ``interp_degree_interpolated_profile`` parameter) to generate a continuously varying
+spatial profile that can be evaluated at any wavelength.[ #173]
+
 API Changes
 ^^^^^^^^^^^
 
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -8,6 +8,7 @@ import numpy as np
 from astropy import units as u
 from astropy.modeling import Model, models, fitting
 from astropy.nddata import NDData, VarianceUncertainty
+from scipy.interpolate import RectBivariateSpline
 
 from specreduce.core import SpecreduceOperation
 from specreduce.tracing import Trace, FlatTrace
@@ -250,6 +251,19 @@ class HorneExtract(SpecreduceOperation):
     Perform a Horne (a.k.a. optimal) extraction on a two-dimensional
     spectrum.
 
+    There are two options for fitting the spatial profile used for
+    extraction - by default, a 1D gaussian is fit and as a uniform profile
+    across the spectrum. Alternativley, the ``self profile`` option may be
+    chosen - when this option is chosen, the spatial profile will be sampled
+    at various locations (set by <>) and interpolated between to produce a
+    smoothly varying spatial profile across the spectrum.
+
+    If using the Gaussian option for the spatial profile, a background profile
+    may be fit (but not subtracted) simultaneously to the data. By default,
+    this is done with a 2nd degree polynomial. If using the
+    ``interpolated_profile`` option, the background model must be set to None.
+
+
     Parameters
     ----------
 
@@ -268,8 +282,24 @@ class HorneExtract(SpecreduceOperation):
         The index of the image's cross-dispersion axis. [default: 0]
 
     bkgrd_prof : `~astropy.modeling.Model` or None, optional
-        A model for the image's background flux.
-        [default: models.Polynomial1D(2)]
+        A model for the image's background flux. If ``spatial_profile`` is set
+        to ``interpolated_profile``, then ``bkgrd_prof`` must be set to None.
+        [default: models.Polynomial1D(2)].
+
+    spatial_profile : str or dict, optional
+        The shape of the object profile. The first option is 'gaussian' to fit
+        a uniform 1D gaussian to the average of pixels in the cross-dispersion
+        direction. The other option is 'interpolated_profile'  - when this
+        option is used, the profile is sampled in bins and these samples are
+        interpolated between to construct a continuously varying, empirical
+        spatial profile for extraction. For this option, if passed in as a
+        string (i.e spatial_profile='interpolated_profile') the default values
+        for the number of bins used (10) and degree of interpolation
+        (linear in x and y, by default) will be used. To set these parameters,
+        pass in a dictionary with the keys 'n_bins_interpolated_profile' (which
+        accepts an integer number of bins) and 'interp_degree' (which accepts an
+        int, or tuple of ints for x and y degree, respectively).
+        [default: gaussian]
 
     variance : `~numpy.ndarray`, optional
         (Only used if ``image`` is not an NDData object.)
@@ -294,6 +324,7 @@ class HorneExtract(SpecreduceOperation):
     image: NDData
     trace_object: Trace
     bkgrd_prof: Model = field(default=models.Polynomial1D(2))
+    spatial_profile: str = 'gaussian'  # can actually be str, dict
     variance: np.ndarray = field(default=None)
     mask: np.ndarray = field(default=None)
     unit: np.ndarray = field(default=None)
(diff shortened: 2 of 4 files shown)


specreduce/extract.py
```diff
@@ -267,9 +281,26 @@ class HorneExtract(SpecreduceOperation):
...
-    bkgrd_prof : `~astropy.modeling.Model`, optional
-        A model for the image's background flux.
-        [default: models.Polynomial1D(2)]
+    bkgrd_prof : `~astropy.modeling.Model` or None, optional
+        A model for the image's background flux. If ``spatial_profile`` is set
+        to ``self_profile``, then ``bkgrnd_prof`` must be set to None.
+        [default: models.Polynomial1D(2)].
+
+    spatial_profile : str, optional
+        The shape of the object profile. Choise of 'gaussian' to fit a uniform
```
kecnry, line 290:
```suggestion
        The shape of the object profile. Choice of 'gaussian' to fit a uniform
```

specreduce/extract.py
```diff
@@ -448,6 +572,22 @@ def __call__(self, image=None, trace_object=None,
         bkgrd_prof : `~astropy.modeling.Model`, optional
             A model for the image's background flux.
 
+        spatial_profile : str, optional
+            The shape of the object profile. Choise of 'gaussian' to fit a uniform
```
kecnry, line 576:
```suggestion
            The shape of the object profile. Choice of 'gaussian' to fit a uniform
```

specreduce/extract.py
```diff
@@ -478,10 +618,22 @@ def __call__(self, image=None, trace_object=None,
...
+                                      interp_degree_self_profile is not None
+                                      else self.interp_degree_self_profile)
         variance = variance if variance is not None else self.variance
         mask = mask if mask is not None else self.mask
         unit = unit if unit is not None else self.unit
 
+        # make sure a valid profile type was specified:
+        spatial_profile = spatial_profile.lower()
+        if spatial_profile not in ['gaussian', 'self_profile']:
+            raise ValueError("spatial_profile must be 'gaussian' or 'self_profile'")
```
kecnry, line 635: Would this help future-proof for when we might add more valid options?

```suggestion
        spatial_profile_choices = ('gaussian', 'self_profile')
        if spatial_profile not in spatial_profile_choices:
            raise ValueError(f"spatial_profile must be one of {', '.join(spatial_profile_choices)}")
```

specreduce/extract.py
```diff
@@ -497,50 +649,79 @@ def __call__(self, image=None, trace_object=None,
...
+                mean_init_guess = trace_object.trace
+            else:
+                mean_init_guess = np.broadcast_to(
+                    img.shape[crossdisp_axis] // 2, img.shape[disp_axis]
+                )
+
+        else:  # self_profile
+            # for now, bkgrnd_prof must be None because a compound model can't
+            # be created with a interpolator + model. i think there is a way
+            # around this, but will follow up later
```
kecnry, line 683: should we create a follow-up issue for this or is it already being tracked somewhere?

specreduce/extract.py
```diff
@@ -497,50 +649,79 @@ def __call__(self, image=None, trace_object=None,
...
+            if bkgrd_prof is not None:
+                raise ValueError('When `spatial_profile`is `self_profile`,'
+                                 '`bkgrnd_prof` must be None. Background should'
+                                 ' be fit and subtracted from `img` beforehand.')
+            # make sure n_bins doesnt exceed the number of (for now) finite
+            # columns. update this when masking is fixed.
+            n_finite_cols = np.logical_or.reduce(or_mask, axis=crossdisp_axis)
+            n_finite_cols = np.count_nonzero(n_finite_cols.astype(int) == 0)
+
+            # determine interpolation degree from input and make
```
kecnry, line 693: and make what?

specreduce/extract.py
```diff
@@ -497,50 +649,79 @@ def __call__(self, image=None, trace_object=None,
...
+            if isinstance(interp_degree_self_profile, int):
+                kx = ky = interp_degree_self_profile
+            else:  # if input is tuple of ints
+                kx, ky = interp_degree_self_profile
```
kecnry, line 697: does there need to be any type-checking (or length checking) here on `interp_degree_self_profile` to avoid cryptic errors?

specreduce/extract.py
```diff
@@ -497,50 +649,79 @@ def __call__(self, image=None, trace_object=None,
...
+
+            if n_bins_self_profile >= n_finite_cols:
+
+                raise ValueError(f'`n_bins_self_profile` ({n_bins_self_profile}) '
+                                 'should not exceed the number of fully-finite '
```
kecnry, line 702: This wording could be confusing for the user when `n_bins_self_profile == n_finite_cols`

```suggestion
                                 'must be less than the number of fully-finite '
```

specreduce/extract.py
```diff
@@ -549,18 +730,32 @@ def __call__(self, image=None, trace_object=None,
...
-            # fit compound model to column
-            fitted_col = fit_ext_kernel(xd_pixels)
-
-            # save result and normalization
-            kernel_vals.append(fitted_col)
-            norms.append(fit_ext_kernel.amplitude_0
-                         * fit_ext_kernel.stddev_0 * np.sqrt(2*np.pi))
+            if self.spatial_profile == 'gaussian':
+
+                # else, set compound model's mean to column's matching trace value
```
kecnry, line 735:
```suggestion
                # set compound model's mean to column's matching trace value
```

specreduce/extract.py
```diff
@@ -549,18 +730,32 @@ def __call__(self, image=None, trace_object=None,
...
+                if bkgrd_prof is not None:  # attr names will be diff. if not compound
+                    fit_ext_kernel.mean_0 = mean_init_guess[col_pix]
+                else:
+                    fit_ext_kernel.mean = mean_init_guess[col_pix]
+
+                # evaluate fit model (with shifted mean, based on trace)
+                fitted_col = fit_ext_kernel(xd_pixels)
+
+                # save result and normalization
+                # this doesn't need to be in this loop, address later
```
kecnry, line 746: is there an issue tracking this?  If not, should we create one so we don't forget?

specreduce/tests/test_extract.py
```diff
@@ -197,3 +239,91 @@ def test_horne_non_flat_trace():
...
+    # option add a perfectly gaussian source and make sure gaussian extraction
+    # and self-profile extraction agree (since self=gaussian in this case)
+
+    image = mk_test_img
+    add_gaussian_source(image)  # add source across image, flat trace
+
+    trace = FlatTrace(image, image.shape[0] // 2)
+
+    print(image.data)
+
```
kecnry, line 268:
```suggestion
```

PatrickOgle (approved):
Performed science review on simulated 2D spectrum with tilted trace and expanding PSF.  This PR gives the proper flux level and improved SNR compared with the existing HorneExtract. SNR improves with profile binning as expected.

PatrickOgle:
Would be useful to expose cross-dispersion profile as an output. (For future PR if not now.)

cshanahan1:
working on rebasing this (which has become frustrating :| ) and addressing kyle's comments, will push changes shortly

cshanahan1:
> Would be useful to give cross-dispersion profile as an output. (For future PR if not now.)

i have it as a private attribute (self._interp_spatial_prof) which i was using for testing, i can just remove the underscore and add a line to the docs about it if thats what you mean?

specreduce/extract.py
cshanahan1: there's no follow up issue - i think we should consider removing this option altogether which is why i didn't spend the time to do this here.

specreduce/extract.py
cshanahan1: lol, its a mystery (just kidding, fixed)

specreduce/extract.py
cshanahan1: added to be safe

specreduce/extract.py
cshanahan1: no, i plan to address this when i address masking. right now it makes sense for it to be in the loop, even though the area under the gaussian doesn't change when the mean shifts (and, we don't need to shift the mean at all since align_along_trace is accounting for this), but it does append NaN to the output array if the column has any masked values which is how masks are treated for now.

eteq:
A few high-level API thoughts:

First, I think the use of keyword arguments that apply only for certain values of another keyword (i.e., `spatial_profile`) is a bit of an anti-pattern.  I see two immediate alternatives, although there are probably other options as well.
1. have `spatial_profile` accept an *object* and then gaussian and self-profile could be classes that take the various keywords as arguments.  This is closer to the idiom used in, say, `photutils` and `specutils` (at least in some places)
2. If that seems a bit too heavy-weight, another option is to use tuples or dicts in the same way.  E.g., `spatial_profile` could take a dict that is something like ``{'name': 'self_profile', 'n_bins_self_profile':5, ...}`` or similar.  That's a bit harder to document and requires some annoying validation logic, but is lighter-weight than a class.

Second, I find the name "self_profile" to be not great because "self" can mean a lot of things. My first thought was "empirical_binned_profile", but that's a bit wordy... maybe someone else has something a bit more compact?

Third, re:
> Should we return the fit profile
from @cshanahan1 - yes, I think so!  I think the user will indeed sometimes want this.

Last, re: background model:  I think the background model is very important for some use cases, but I see the complexity.  FWIW, this is yet another example of the problem that astropy modeling really needs an implementation of splines... but I think it's ok to make an issue that basically says that's the core of the problem, and then whenever that's possible, we'll add background support here, but otherwise not prioritizing it. Realistically, I think the empirical nature of this fit means the cases where background subtraction is important (strong skylines) is better suited to the gaussian case anyway.

kecnry:
@eteq - if we want different input arguments, different exposed methods, and different defaults for background behavior.... it's feeling more and more to me like we should just have them as separate classes (likely with shared code, consistent APIs, and not necessarily needing to be split in this PR).

cshanahan1:
Thanks for the feedback @eteq! What do you think about swapping to 'empirical_profile', I think that is a bit more informative.

Though its outside of the scope of this current ticket, there will eventually be at least a third independent option for spatial profile (that also won't share the 2 new keywords here, and may have its own). Keeping that in mind, it seems there are a few options for how to organize this

1. (What is currently in this PR) a single HorneExtract class, with an argument ('spatial_profile') that dictates which type will be used ('gaussian' and 'self_profile' for now, perhaps a 'user_supplied' option later). For each type, there are a set of keyword arguments that will not be used for other types (i.e. n_bins_self_profile is ignored if spatial_profile is 'gaussian', but does not throw an error if set). 

2. Also just one HorneExtract class, but 'spatial_profile' accepts a dictionary if there are associated keyword args or just a string if not. For example, since gaussian doesn't require any additional info you would set it by doing spatial_profile='gaussian', but to use the self profile option you could either do spatial_profile='self_profile' to use the default for n_bins and interp_degree, or spatial_profile={'method':'self_profile', 'n_bins_self_profile':50}. This has the benefit of only having one arg that controls the spatial profile, but would be more annoying to just quickly change the number of bins from the default since you'd have to change your input from a string to a dict and remember what keys it wants etc.

3. A single HorneExtract class, but with different 'call' methods for each extraction type. Constructing the Horne class would not take any of the spatial_profile arguments when created, those would be supplied to each HorneExtract().gaussian_profile_extraction, HorneExtract().empirical_profile_extraction()

4. Multiple HorneExtract classes, for each profile type.

Any other ideas? Or votes? I am leaning towards option 2.

eteq:
`empirical_profile` is definitely better, although I think it's still not ideal since there's still several "empirical" things. But I don't have an obvious better suggestion...

Also, in a side-conversation, @kecnry @tepickering @rosteen and @camipacifici discussed the interface question and I think we settled on option 2 *for now* to finish this PR, but that we need to come to consensus on solving the problem here in a consistent way across specreduce, and we can re-jigger this to fit that once that decision is made.

tepickering:
agreed, option 2 _for now_ is fine. ultimately, both `spatial_profile` and `background_profile` should have the same signature: `callable(wavelength) -> 1D array | callable(spatial)`. i think it's fine for `HorneExtract` to support simple defaults if those aren't specified, but real world data can be highly non-rectilinear and the design will need to handle that eventually.

CHANGES.rst
```diff
@@ -4,6 +4,13 @@
 New Features
 ^^^^^^^^^^^^
 
+- Added 'self_profile' option for HorneExtract. If The ``self_profile`` option
```
tepickering, line 7: i think `interpolated_profile` is a better name and more accurately descriptive of it.

specreduce/extract.py
```diff
@@ -250,6 +251,19 @@ class HorneExtract(SpecreduceOperation):
...
     spectrum.
 
+    There are two options for fitting the spatial profile used for
+    extraction - by default, a 1D gaussian is fit and as a uniform profile
+    across the spectrum. Alternativley, the ``self profile`` option may be
+    chosen - when this option is chosen, the spatial profile will be sampled
+    at various locations (set by <>) and interpolated between to produce a
+    smoothly varying spatial profile across the spectrum.
+
+    If using the Gaussian optoin for the spatial profile, a background profile
```
tepickering, line 261: optoin -> option

specreduce/extract.py
```diff
@@ -294,6 +325,9 @@ class HorneExtract(SpecreduceOperation):
     image: NDData
     trace_object: Trace
     bkgrd_prof: Model = field(default=models.Polynomial1D(2))
```
tepickering, line 327: i would rename this `background_profile` for consistency with `spatial_profile`.

specreduce/tests/test_extract.py
```diff
@@ -14,17 +15,49 @@
...
+@pytest.fixture
+def mk_test_img(nrows=30, ncols=10):
+    image = np.ones(shape=(nrows, ncols))
+    for j in range(image.shape[0]):
+        image[j, ::] *= j
+    image = CCDData(image, unit=u.Jy)
+    return image
+
+
+def add_gaussian_source(image, amps=2, stddevs=2, means=None):
```
tepickering, line 29: this reminds me that i need to refactor `specreduce.synth_data.make_2d_trace_image` to have `profile` as an input parameter. right now it's hard-coded to `Moffat1D`.

tepickering (changes requested):
i do have a fundamental concern that building an interpolated spatial profile as a function of wavelength is outside the scope of a class called `HorneExtract`. the implementation here also implicitly assumes rectilinear data which is rarely the case in the real world. 

that said, this is a good first draft of the functionality and something that can be built on. i would prefer the use of `interpolated_profile` vs `self_profile`. `empirical_profile` as @eteq suggested also works. (ultimately, it should just be `spatial_profile` and should be treated the same internally regardless of model.) i would also address the lines that `codecov` catches as not covered by tests.

CHANGES.rst
cshanahan1: done

specreduce/extract.py
cshanahan1: should we do this in another PR so it has its own change log entry for clarity since this changes the interface?

tepickering: yeah, that's appropriate.

cshanahan1:
@tepickering @kecnry @PatrickOgle @eteq any objections to merging? i think i addressed all comments and i also just added some more test coverage

tepickering (approved):

tepickering:
thanks for all this work! changes look good to me and the checks all pass now so merge away.

PR #175, opened 2023-05-22
Expose residuals in WavelengthCalibration1D

cshanahan1:
Add ``fit_residuals`` attribute to ``WavelengthCalibration1D``. These are computed by evaluating the model after the fit is performed, which is done when ``apply_to_spectrum`` is called.

(Closes JDAT-3405)

### DIFF
--- a/specreduce/tests/test_wavelength_calibration.py
+++ b/specreduce/tests/test_wavelength_calibration.py
@@ -1,4 +1,5 @@
 from numpy.testing import assert_allclose
+import numpy as np
 import pytest
 
 from astropy.table import QTable
@@ -83,3 +84,30 @@ def test_expected_errors(spec1d):
 
     with pytest.raises(ValueError, match="specify at least one"):
         WavelengthCalibration1D(spec1d, line_pixels=centers)
+
+
+def test_fit_residuals(spec1d):
+    # test that fit residuals are all 0 when input is perfectly linear and model
+    # is a linear model
+
+    centers = np.array([0, 10, 20, 30])
+    w = (0.5 * centers + 2) * u.AA
+    test = WavelengthCalibration1D(spec1d, line_pixels=centers,
+                                   line_wavelengths=w)
+
+    test.apply_to_spectrum(spec1d)  # have to apply for residuals to be computed
+
+    assert_quantity_allclose(test.fit_residuals, 0.*u.AA, atol=1e-07*u.AA)
+
+
+def test_fit_residuals_access(spec1d):
+    # make sure that accessing fit_residuals fails if .wcs hasn't been accessed
+    # yet, which is the property that performs the fit. ``apply_to_spectrum``
+    # also accesses .wcs, so if this is run fit_residuals will be available as well
+
+    centers = np.array([0, 10, 20, 30])
+    w = (0.5 * centers + 2) * u.AA
+    test = WavelengthCalibration1D(spec1d, line_pixels=centers,
+                                   line_wavelengths=w)
+    with pytest.raises(ValueError):
+        test.fit_residuals
--- a/specreduce/tests/test_wavelength_calibration.py
+++ b/specreduce/tests/test_wavelength_calibration.py
@@ -109,5 +109,8 @@ def test_fit_residuals_access(spec1d):
     w = (0.5 * centers + 2) * u.AA
     test = WavelengthCalibration1D(spec1d, line_pixels=centers,
                                    line_wavelengths=w)
-    with pytest.raises(ValueError):
+    expected_msg = ('Fit residuals are only available after the new WCS is'
+                    ' fit - this can be done by accessing the ``.wcs`` attribute,'
+                    ' or by calling the ``.apply_to_spectrum`` method.')
+    with pytest.raises(ValueError, match=expected_msg):
         test.fit_residuals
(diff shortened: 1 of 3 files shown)


specreduce/wavelength_calibration.py
```diff
@@ -191,6 +191,10 @@ def model(self, new_model):
         self._clear_cache()
         self._model = new_model
 
+    @property
+    def fit_residuals(self):
+        return self._fit_resids
```
kecnry, line 196: should `_fit_resids` be initialized to `None` in the `__init__` to avoid an error when accessing this before calling `apply_to_spectrum`?  Should that case raise a useful error stating that `apply_to_spectrum` needs to be called first?

specreduce/wavelength_calibration.py
```diff
@@ -209,6 +213,9 @@ def wcs(self):
         # Fit the model
         self._model = fitter(self._model, x, y)
 
+        # Get the fit residuals by evaulating model
+        self._fit_resids = y - self._model(x)
```
kecnry, line 217: I wonder if this should just be its own cached property so that it knows to clear when the model is manually changed by the user and to save a (small) bit of time when applying to a spectrum but not requesting residuals?

specreduce/wavelength_calibration.py
```diff
@@ -191,6 +191,10 @@ def model(self, new_model):
         self._clear_cache()
         self._model = new_model
 
+    @property
+    def fit_residuals(self):
```
kecnry, line 195: probably should have a docstring to explain what this is and that its only available after calling `apply_to_spectrum`.

specreduce/wavelength_calibration.py
cshanahan1: My thought was that if it was not initialized, it WOULD raise an error if apply_to_spectrum wasn't run first.

I could do it this way - initialize it to None, have a check in the getter that if fit_residuals is None, raise a more informative error. Do you prefer that approach?

kecnry: Yes, but the error will seem cryptic to the user (just something like "WavelengthCalibration object has no attribute _fit_resids").  I think either what you suggest or the cached_property could both catch the error and suggest to the user what needs to be done from their end.

specreduce/wavelength_calibration.py
cshanahan1: done

specreduce/wavelength_calibration.py
cshanahan1: done

specreduce/wavelength_calibration.py
cshanahan1: oh actually, it would be accessible after calling either .wcs or .apply_to_spectrum. the fit is done in .wcs.

specreduce/wavelength_calibration.py
cshanahan1: i also added a test to make sure this error is raised when this happens

kecnry: would checking for `self._model is None` for the error get around the need to set `_fit_resids` to False?

specreduce/tests/test_wavelength_calibration.py
```diff
@@ -83,3 +84,30 @@ def test_expected_errors(spec1d):
...
+def test_fit_residuals_access(spec1d):
+    # make sure that accessing fit_residuals fails if .wcs hasn't been accessed
+    # yet, which is the property that performs the fit. ``apply_to_spectrum``
+    # also accesses .wcs, so if this is run fit_residuals will be available as well
+
+    centers = np.array([0, 10, 20, 30])
+    w = (0.5 * centers + 2) * u.AA
+    test = WavelengthCalibration1D(spec1d, line_pixels=centers,
+                                   line_wavelengths=w)
+    with pytest.raises(ValueError):
```
rosteen, line 112: I think we generally use `match="some text"` in the `pytest.raises` call to make sure that the specific error being raised is the one we expect, in case some other error starts happening at some point.

rosteen (commented):
One minor suggestion to use `match` in the error test, other than that it looks good. I don't love the `None` vs `False` distinction to check to see if the residuals are available, but I also don't know that there's a better way to do it...can't check to see if `wcs` is `None` because that would just force the model to fit 😅 .

specreduce/wavelength_calibration.py
cshanahan1: no, because both ._model and .model will always exist (default is a linear model), just not fit if .wcs isn't accessed.

specreduce/tests/test_wavelength_calibration.py
cshanahan1: done!

rosteen (approved):
LGTM, thanks!

cshanahan1:
and yeah, regarding comments on how to raise an error when fit_residuals is accessed with .wcs never being run - we can't check if _fit_resids is None (which is what its initialized to) because that would also fail upon the first time computing residuals. there needs to be some switch flipped by .wcs to indicate that residuals can now be fit, so i can either do what i did or make a new private attribute

kecnry (approved):
As discussed offline, avoiding the `_fit_resids = False` and needing to call `wcs` or `apply_to_spectrum` _before_ `fit_residuals` will likely result in a slight change in behavior to how successful calls will act, so I'll open that as a follow-up PR for discussion after this is merged (since this accomplishes the immediate goals).  Thanks @cshanahan1!

cshanahan1:
im gonna merge this now since it has two approvals. thanks!

PR #176, opened 2023-05-24
store the fitted_model as its own cached_property

kecnry:
As a follow-up to #175, this refactors the code slightly so that it is not necessary to call `cal.wcs` or `cal.appy_to_spectrum` before calling `cal.residuals` (and also does some renaming from `fit_residuals` > `residuals` and `model` > `input_model`).  This also separates the _input_ `cal.input_model` from the _output_ `cal.fitted_model` and therefore slightly changes the behavior if calling successive fits on the same calibration object.  If you want to "pick up where you left off", you could now do something like:

```
cal = WavelengthCalibration1D(lamp_spectrum_1, model=Linear1D, ...)
cal.apply_to_spectrum(science_spectrum_1)

cal.input_spectrum = lamp_spectrum_2
cal.input_model = cal.fitted_model
cal.apply_to_spectrum(science_spectrum_2)
```

### DIFF
--- a/specreduce/wavelength_calibration.py
+++ b/specreduce/wavelength_calibration.py
@@ -213,10 +213,7 @@ class WavelengthCalibration1D():
     def residuals(self):
         """
         calculate fit residuals between matched line list pixel centers and
-        wavelengths and the evaluated fit model. this is only accessible
-        after the ``wcs`` attribute is accessed, where the fit is actually
-        performed. ``apply_to_spectrum`` also calls .wcs and therefore
-        performs the fit, so this is also acessible if that is called.
+        wavelengths and the evaluated fit model.
         """
 
         x = self._matched_line_list["pixel_center"]
--- a/specreduce/wavelength_calibration.py
+++ b/specreduce/wavelength_calibration.py
@@ -65,11 +65,10 @@ class WavelengthCalibration1D():
         """
         self._input_spectrum = input_spectrum
         self._model = model
-        self._cached_properties = ['wcs', 'fit_residuals',]
+        self._cached_properties = ['fit_model', 'fit_residuals', 'wcs']
         self.fitter = fitter
         self._potential_wavelengths = None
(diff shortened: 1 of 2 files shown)


cshanahan1:
I haven't tested this out yet but the concept makes more sense to me. I think people may assume that it is going to refit your original input to-be-fit model (which it wasn't doing before), but having the option to also re-fit a previously fit model should also be an option. 

There should be a test that covers running the fit multiple time and makes sure it works as intended.

Also, a larger question that has been in my mind since I've started working on specreduce - is there any planned effort to write thorough documentation? As it stands there isn't even a section in the documentation to make note of this changed behavior. There's only the Getting Started guide.

cshanahan1:
maybe having 'input_model' and 'fit_model' is clearer?

kecnry:
> maybe having 'input_model' and 'fit_model' is clearer?

@rosteen - any thoughts?

(I also see that this broke the recent tests - I'll mark as draft until we can get those updated)

specreduce/wavelength_calibration.py
```diff
@@ -184,67 +183,59 @@ def input_spectrum(self, new_spectrum):
...
             fitter = self.fitter
 
         # Fit the model
-        self._model = fitter(self._model, x, y)
+        return fitter(self.input_model, x, y)
+
+    @cached_property
+    def residuals(self):
+        """
+        calculate fit residuals between matched line list pixel centers and
```
cshanahan1, line 215: this docstring is no longer true since accessing this will cause the model to be fit. i think you could just leave this at "calculate fit residuals between matched line list pixel centers and wavelengths and the evaluated fit model."

cshanahan1 (changes requested):
this looks good, I just left one comment about an outdated docstring

cshanahan1 (approved):

rosteen (approved):
Looks fine to me, no objections.

PR #177, opened 2023-06-01
enforce spectral axis being left to right for wavelength calibration

cshanahan1:


specreduce/wavelength_calibration.py
```diff
@@ -70,6 +70,15 @@ def __init__(self, input_spectrum, matched_line_list=None, line_pixels=None,
...
         self._catalog = catalog
 
+        if not isinstance(input_spectrum, Spectrum1D):
+            raise ValueError('Input spectrum must be Spectrum1D.')
+
+        # make sure spec. is dispersed right to left, should be user's responsibility
+        spec_axis = input_spectrum.spectral_axis.value
+        sorted_increasing = np.all(spec_axis[1:] >= spec_axis[:-1])
+        if not sorted_increasing:
+            raise ValueError('Spectrum must be dispersed left to right.')
```
rosteen, line 80:
```suggestion
            raise ValueError('Spectral axis values must be increasing from left to right.')
```
I think "dispersed left to right" is potentially ambiguous, may as well say explicitly what we're checking for.

specreduce/wavelength_calibration.py
```diff
@@ -115,15 +129,20 @@ def __init__(self, input_spectrum, matched_line_list=None, line_pixels=None,
...
             if isinstance(line_wavelengths, u.Quantity):
-                # Ensure frequency is descending or wavelength is ascending
-                if str(line_wavelengths.unit.physical_type) == "frequency":
-                    line_wavelengths[::-1].sort()
-                else:
-                    line_wavelengths.sort()
+                if not _check_arr_monotonic(line_wavelengths):
+                    if str(line_wavelengths.unit.physical_type) == "frequency":
+                        raise ValueError('Frequencies must be strictly increasing or decreasing.')
+                    raise ValueError('Wavelengths must be strictly increasing or decreasing.')
```
kecnry, line 139: Not for this PR, but I wonder if we should rename `line_wavlengths` to `line_spectral_values` or something?

specreduce/wavelength_calibration.py
```diff
@@ -115,15 +129,20 @@ def __init__(self, input_spectrum, matched_line_list=None, line_pixels=None,
...
+
                 self._matched_line_list["wavelength"] = line_wavelengths
+
             elif isinstance(line_wavelengths, QTable):
-                line_wavelengths.sort("wavelength")
+                if not _check_arr_monotonic(line_wavelengths['wavelength']):
+                    raise ValueError('Wavelengths must be strictly increasing or decreasing.')
```
kecnry, line 145: do we support a frequency option for the QTable case?

specreduce/wavelength_calibration.py
```diff
@@ -27,6 +27,17 @@ def concatenate_catalogs():
...
 
+def _check_arr_monotonic(arr):
+    # returns True if ``arr`` is either strictly increasing or strictly
+    # decreasing, otherwise returns False
+
+    sorted_increasing = np.all(arr[1:] >= arr[:-1])
+    sorted_decreasing = np.all(arr[1:] <= arr[:-1])
+    if sorted_increasing or sorted_decreasing:
+        return True
+    return False
```
kecnry, line 38:
```suggestion
    return sorted_increasing or sorted_decreasing
```

specreduce/tests/test_wavelength_calibration.py
```diff
@@ -109,3 +109,25 @@ def test_fit_residuals_access(spec1d):
                                    line_wavelengths=w)
     test.residuals
     test.wcs
+
+
+def test_unsorted_pixels_wavelengths(spec1d):
```
kecnry, line 114: can we also add a test for the inverted case to make sure that does not pass?  Perhaps `test_linear_from_list` can just have a second case that does `WavelengthCalibration1D(spec1d, line_pixels=centers[::-1], line_wavelengths=w)`.

kecnry (commented):
Just a few small comments/suggestions, but otherwise I think this looks good and is an elegant solution.  Thanks!

specreduce/wavelength_calibration.py
cshanahan1: yeah, also we should not be checking strictly for a 'wavelength' column if frequencies are allowed.

specreduce/wavelength_calibration.py
cshanahan1: thats what was implied, but are checks for the presence of a 'wavelength' column so it seems like you could input something with frequency units but the column must be named wavelength which is confusing.

i also am not a fan of all the different input options available - i feel like it would be a lot less confusing if we just forced users to input a table with pixel and wavelength or frequency (or eventually, the catalog option) instead of allowing two matched arrays, or a table, or a table with a pixel column and an array, or two tables each with one col. i think this is too many options and a lot of this code is just figuring out the logic of the inputs where we could just insist on users constructing a table beforehand (or, and i might like this option better because then youre not checking for table column names, an input pixel array and an input wavelength/freq array, and then internally convert to a table).

kecnry (approved):

rosteen (approved):
LGTM, and I agree that a separate effort to simplify the inputs would be worth doing. I erred too far on the side of flexibility when I first implemented the class.

PR #182, opened 2023-06-21
Initial implementation of line matching

tepickering:
This PR adds functions to find and centroid lines in a calibration, e.g. arc lamp, spectrum and then match pixel positions to wavelengths using an input WCS. More work is needed to close the loop between this and the fitting process for wavelength calibration, but this is a start.

This PR also contains come code cleanups and the addition of typing in several places. To take advantage of the significant improvements in typing and type handling in newer versions of python, the minimum python version has been upped to 3.10. The CI and `tox` configuration has been updated to reflect this.


specreduce/line_matching.py
```diff
@@ -0,0 +1,121 @@
...
+
+    window : The window size in units of fwhm to use for the gaussian fits.
+
+    noise_factor : The factor to multiply the uncertainty by to determine the noise threshold
+        in the `~specutils.fitting.find_lines_threshold` routine.
+
+    Returns
+    -------
+    QTable
+        A table of detected arc lines and their properties.
```
cshanahan1, line 48: Describe what table columns are returned (seems like just pixel_center and wavelength?)

specreduce/utils/synth_data.py
```diff
@@ -1,11 +1,13 @@
...
 
+from typing import Sequence
+
 import numpy as np
 
 from photutils.datasets import apply_poisson_noise
 
 import astropy.units as u
-from astropy.modeling import models
+from astropy.modeling import Model, models
```
cshanahan1, line 10: sorry I'm a bit confused about this - you're importing 'Model' but also redefining 'Model' below.

specreduce/tests/test_line_matching.py
```diff
@@ -0,0 +1,110 @@
...
+@pytest.mark.remote_data
+@pytest.mark.filterwarnings("ignore:No observer defined on WCS")
+@pytest.mark.filterwarnings("ignore:Model is linear in parameters")
+def test_find_arc_lines(mk_test_data):
+    """
+    Test the find_arc_lines routine.
+    """
+    _, _, _, arc_sub = mk_test_data
+    lines = find_arc_lines(arc_sub, fwhm=5, window=5, noise_factor=5)
+    assert len(lines) > 1
```
cshanahan1, line 92: maybe both of these tests should check that the correct thing was returned rather than just a non-empty table?

specreduce/utils/synth_data.py
tepickering: i am importing `Model` to use as a type declaration and where it looks like i'm redefining it, i am actually defining the default `Model` to use as input.

specreduce/tests/test_line_matching.py
tepickering: my goal here was to make sure all lines of code run through without error and at least one line is detected. validating inputs vs. outputs is, imho, the next level of development because it'll depend on the details of how lines are centroided.

tepickering:
i think i successfully merged in the significant changes from #202. the python 3.8 and 3.9 tests did fail as expected so i bumped `requires-python` to 3.10 as we agreed to initially. i tweaked `tox.ini` and the workflows accordingly to get the tests to all pass.

tepickering:
this PR has been languishing for a while. i'd like to finally get this in to at least add the type annotation work and up the minimum python to 3.10. there's more work to be done on the line matching and especially wavelength calibration, but at this point it's better off done in new PRs.

specreduce/line_matching.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Extra sanity handling to make sure the input Sequence can be converted to an np.array
+    try:
+        pixel_positions = np.array(pixel_positions, dtype=float)
+    except ValueError as e:
+        raise ValueError(f"pixel_positions must be convertable to np.array with dtype=float: {e}")
+
+    catalog_pixels = spectral_wcs.world_to_pixel(catalog_wavelengths)
+    separations = pixel_positions[:, np.newaxis] - catalog_pixels
+    matched_loc = np.where(np.abs(separations) < tolerance)
```
cshanahan1, line 117: maybe raise a warning when there are no matches? it might be more informative than just returning an empty table

specreduce/tests/test_line_matching.py
```diff
@@ -0,0 +1,110 @@
+import pytest
```
cshanahan1, line 1: alphabetize imports

specreduce/tests/test_line_matching.py
```diff
@@ -0,0 +1,110 @@
...
+        'CRPIX1': 519.8,       # Reference pixel [pix]
+        'CRVAL1': 7245.2,      # Reference value [Angstrom]
+        'CDELT1': 2.956,       # Linear dispersion [Angstrom/pix]
+        'CTYPE2': 'PIXEL',     # Spatial detector coordinates
+        'CUNIT2': 'pix',       # Spatial units
+        'CRPIX2': 1,           # Reference pixel
+        'CRVAL2': 0,           # Reference value
+        'CDELT2': 1            # Spatial units per pixel
+    }
+    non_linear_wcs = WCS(header=non_linear_header)
```
cshanahan1, line 54: does it make sense to also test this with GWCS since the function can accept either? im not sure if thats overkill

cshanahan1 (commented):
still testing out the functionality, but some initial comments on code.

I noticed that throughout, you've removed some of the numpy style docstrings that describe input parameter type and default, and are instead annotating the function. I think that the annotations are a helpful addition, but that the docstrings should retain the same format as before, making sure defaults are also described.

cshanahan1:
> > still testing out the functionality, but some initial comments on code.
> > I noticed that throughout, you've removed some of the numpy style docstrings that describe input parameter type and default, and are instead annotating the function. I think that the annotations are a helpful addition, but that the docstrings should retain the same format as before, making sure defaults are also described.
> 
> this was a very intentional change. now that type annotations are supported, they should be the preferred way to document typing and defaults. it is a much more flexible and powerful way of doing so. e.g., it empowers compilers to optimize the code using the given information which is important in contexts such as GPUs. what we definitely DO NOT WANT is to have redundant information in two places because it is guaranteed to create inconsistencies in the future.

That makes sense not to keep in in two places. So this is the new best practice then? If so maybe we can make a PR to do this across the whole package so its not in the diff here, that should be fairly quick to do

specreduce/__init__.py
```diff
@@ -1,10 +1,10 @@
...
 
-# Packages may add whatever they like to this file, but
-# should keep this content at the top.
-# ----------------------------------------------------------------------------
-from ._astropy_init import *   # noqa
-# ----------------------------------------------------------------------------
-
 from specreduce.core import *  # noqa
 from specreduce.wavelength_calibration import * # noqa
+
```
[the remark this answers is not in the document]
tepickering: it may not be. i just moved it over from `_astropy_init.py` as-is so there'd be no surprises.

specreduce/line_matching.py
```diff
@@ -0,0 +1,121 @@
...
+    Returns
+    -------
+    QTable
+        A table of the matched lines and their pixel/wavelength positions.
+    """
+
+    # This routine uses numpy broadcasting which doesn't always behave with Quantity objects.
+    # Convert to u.pix and pull out the np.ndarray values.
+    if isinstance(pixel_positions, u.Quantity):
+        pixel_positions = pixel_positions.to(u.pix).value
```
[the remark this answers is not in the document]
tepickering: yeah, safest to just strip the unit in this case.

specreduce/line_matching.py
```diff
@@ -0,0 +1,121 @@
...
+        fwhm *= u.pix
+
+    detected_lines = find_lines_threshold(spectrum, noise_factor=noise_factor)
+    detected_lines = detected_lines[detected_lines['line_type'] == 'emission']
+
+    centroids = []
+    widths = []
+    amplitudes = []
+    for r in detected_lines:
+        g_init = models.Gaussian1D(
```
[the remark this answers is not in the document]
tepickering: it's units-aware and is ultimately getting the unit information from the `Spectrum1D` instance. i was, however, incorrect to assume `fwhm` should be pixels if not specified. it should be in whatever the `spectral_axis` unit is in the input spectrum.

cshanahan1 (approved):
Tests are passing so I'm going to merge this, thanks!

PR #193, opened 2023-11-15
Implement 2D sky subtraction using bivariate splines

tepickering:
This PR implements the [Kelson (2003)](https://www.jstor.org/stable/10.1086/375502) technique for subtracting the sky background from 2D spectra without the need for rectification or interpolation.


chris-simpson:
I really don't think this is an implementation of Kelson's method. You're fitting the BivariateSpline as a function of IMAGE pixel coordinates

```spline_fit = LSQBivariateSpline(x.ravel(), y.ravel(), ...)```

whereas Kelson says the spline should be fit as a function of x_r and y_r (page 694), coordinates which are orthogonal and aligned with the spatial and dispersion axes, using the known transformation from measuring the distortion from arc lamps or sky lines (page 691).

PR #194, opened 2023-11-22
fix warning message in FitTrace when many bins masked

cshanahan1:
This PR consolidates warning messages that were previously produced for each individual masked column into one message reporting all masked columns, or that many columns were masked if there are more than 20.

### DIFF
--- a/specreduce/tracing.py
+++ b/specreduce/tracing.py
@@ -300,14 +300,14 @@ class FitTrace(Trace, _ImageParser):
                              self.bins + 1, dtype=int)
         y_bins = np.tile(np.nan, self.bins)
 
+        warn_bins = []
         for i in range(self.bins):
             # repeat earlier steps to create gaussian fit for each bin
             z_i = img[ilum2, x_bins[i]:x_bins[i+1]].sum(axis=self._disp_axis)
             if not z_i.mask.all():
                 peak_y_i = ilum2[z_i.argmax()]
             else:
-                warnings.warn(f"All pixels in bin {i} are masked. Falling "
-                              'to trace value from all-bin fit.')
+                warn_bins.append(i)
                 peak_y_i = peak_y
 
             if self.peak_method == 'gaussian':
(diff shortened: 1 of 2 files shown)


specreduce/tracing.py
```diff
@@ -346,6 +346,16 @@ def __post_init__(self):
                 # TODO: implement smoothing with provided width
                 y_bins[i] = ilum2[z_i.argmax()]
 
+        # warn about fully-masked bins (which, currently, means any bin with a single masked value)
+        if len(warn_bins) > 0:
+            plural = 's' if len(warn_bins) > 1 else ''
+            # if there are a ton of bins, we don't want to print them all out
+            print(warn_bins)
```
kecnry, line 353:
```suggestion
```

specreduce/tracing.py
```diff
@@ -346,6 +346,15 @@ def __post_init__(self):
...
                 y_bins[i] = ilum2[z_i.argmax()]
 
+        # warn about fully-masked bins (which, currently, means any bin with a single masked value)
+        if len(warn_bins) > 0:
+            plural = 's' if len(warn_bins) > 1 else ''
+            # if there are a ton of bins, we don't want to print them all out
+            if len(warn_bins) > 20:
+                warn_bins = warn_bins[0: 10] + ['...'] + [warn_bins[-1]]
+            warnings.warn(f"All pixels in bin{plural} {', '.join([str(x) for x in warn_bins])}"
+                          " are masked. Falling to trace value from all-bin fit.")
```
kecnry, line 356: I know this was the previous message, but maybe we should try to clarify while we're editing this - is this more accurate?  (If so, the test would need updating as well).

```suggestion
                          " are masked. Falling back on trace value from all-bin fit.")
```

specreduce/tracing.py
```diff
@@ -346,6 +346,15 @@ def __post_init__(self):
                 # TODO: implement smoothing with provided width
                 y_bins[i] = ilum2[z_i.argmax()]
 
+        # warn about fully-masked bins (which, currently, means any bin with a single masked value)
+        if len(warn_bins) > 0:
+            plural = 's' if len(warn_bins) > 1 else ''
+            # if there are a ton of bins, we don't want to print them all out
+            if len(warn_bins) > 20:
+                warn_bins = warn_bins[0: 10] + ['...'] + [warn_bins[-1]]
+            warnings.warn(f"All pixels in bin{plural} {', '.join([str(x) for x in warn_bins])}"
```
kecnry, line 355: Could do the `plural` logic inline (and remove it from earlier)... might be more readable, but up to you.

```suggestion
            warnings.warn(f"All pixels in {'bins' if len(warn_bins) else 'bin'} {', '.join([str(x) for x in warn_bins])}"
```

kecnry (commented):
Just a few small suggestions, but otherwise looks good to me once this is rebased to handle the conflicts.  Thanks!

specreduce/tracing.py
cshanahan1: done

kecnry (approved):

PR #196, opened 2023-11-30
Document with terms discussed during the recent NOIRLab workshop

kbwestfall:
This is to start a living doc where we hold terminology related to development of spectroscopic data reduction and analysis tools.  This is a straight port of the notes from the NOIRLab workshop into an rst doc.  The workshop generally thought it would be useful to have this be a living document, and I thought it was better suited to a versioned doc that gets displayed along with the code documentation (as opposed to a random Google doc).


kbwestfall:
Hi @tepickering .  Just pinging you for a review of this, when you have time.

tepickering (approved):
thanks for this, @kbwestfall! sorry to be slow in merging it.

PR #202, opened 2023-12-05
MNT: Infrastructure and other updates

### LINKED ISSUE #37: Review the testing coverage
For right now, we have stripped down the tests to what is required for consistency checking and testing the package.   As we approach a release, we should increate the tests to test against different versions of astropy and numpy, different OS, and any other tests necessary for full coverage

### LINKED ISSUE #133: synphot should be an OPTIONAL dependency
It is only used for one particular specialized parser and only as fallback. Pulling in the package on install is totally overkill, especially given that synphot has C-extension (which would cause your user to install from source if I happen not to distribute wheel for some exotic distro).

https://github.com/astropy/specreduce/blob/7e7e4a6f73089fb2ec3bcb1d65985cdd8eacb2c9/setup.cfg#L22

https://github.com/astropy/specreduce/blob/7e7e4a6f73089fb2ec3bcb1d65985cdd8eacb2c9/specreduce/calibration_data.py#L179-L187

### LINKED ISSUE #186: TST: Weekly tests are broken
I think it is just a matter of cleaning up `tox.ini`?

https://github.com/astropy/specreduce/actions/runs/6218799805/job/16875792303

### LINKED ISSUE #189: Specreduce layout inconsistent with Photutils and Specutils
I recommend having a similar formatting, along with a specific logo, consistent with Photutils and Specutils.

### LINKED ISSUE #191: Turn unhandled warnings in CI into errors
Even with #190 , there are two deprecation warnings in the devdeps job. If it is from upstream and has been patched upstream, should pull either the dev version of upstream, or ignore it.

### LINKED ISSUE #198: Modernize packaging infrastructure
The python community is moving away from the use of `setup.py` and `setup.cfg` towards the use of `pyproject.toml` to define all configuration and metadata. See [PEP 621](https://peps.python.org/pep-0621/) and the [PyPA Specs page](https://packaging.python.org/en/latest/specifications/) for details.

pllim:
I stared into the code and I couldn't unsee things, so here it is.

* Fix #37 (you already have coverage so just adding a badge)
* Fix #133 (while we're at this, you don't really need photutils or matplotlib at runtime either)
* Fix #186 
* Fix #189
* Fix #191 (was mostly there)
* Fix #198 (you want modernization, you are getting modernization)
* Apply astropy/astropy#15638
* Other fixes, especially tests that were not set up properly.
* Also looks like you have remote data tests but never used `--remote-data` flag in any of your CI job.
* And you also download a remote file blindly for tests not marked as remote data. I turned that file into a proper package test data that is local so you would stop spamming Box.

~Unable to handle this one until I get access:~

* #203 

After merge, can add new job names to branch protection.


pllim:
Close/reopen to see if RTD pops up.

specreduce/fluxcal.py
```diff
@@ -260,6 +257,7 @@ def standard_sensfunc(self, standard, mode='linear', polydeg=9,
         sensfunc_spec = Spectrum1D(spectral_axis=obj_wave, flux=sensfunc_out)
 
         if display is True:
```
tepickering, line 259: should modify or add test to make sure this block gets covered. doesn't look like it has been. an issue can be filed to make sure this gets followed up on.

.github/workflows/tox-tests.yml
```diff
@@ -11,64 +11,44 @@ on:
...
+        - name: Codestyle
+          linux: codestyle
+
+        - name: Python 3.8 on Linux with oldest supported dependencies
+          linux: py38-test-alldeps-oldestdeps
+          toxargs: -v
+
+        - name: Python 3.9 on Windows with minimal dependencies
+          windows: py39-test
+          toxargs: -v
```
tepickering, line 40: we have discussed and decided to make 3.10 the minimum supported version. alas, that change is still staged in #182 which hasn't been merged yet.

tepickering (approved):
this is so awesome! so many cleanups and fixes i've wanted to see, but never got around to. the migration to `pyproject.toml` is also extremely welcome. my only niggle is that we made the decision on slack 7 months ago to set `python_requires` to `>=3.10`. this was implemented in #182, but not yet merged into `main`. i'm fine with merging this as-is and rebasing/revising #182 accordingly. otoh, this is probably the more sensible PR to make that change given the rest of the changes its making to package config/infrastructure.

tox.ini
```diff
@@ -1,21 +1,16 @@
 [tox]
 envlist =
-    py{310,311,312}-test{,-devdeps,-predeps}{,-cov}
-    build_docs
+    py{38,39,310,311,312}-test{,-alldeps}{,-oldestdeps,-devdeps,-predeps}{,-cov}
```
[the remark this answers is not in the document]
pllim: Even numpy has not dropped Python 3.9 yet. Are you sure you want to do it so soon?

https://github.com/numpy/numpy/blob/c9340c210666afefb6946fa6f3a105856ff0e418/pyproject.toml#L19

specreduce/fluxcal.py
pllim: I feel like a display test is going to be very involved, so I opened https://github.com/astropy/specreduce/issues/204 instead, as requested.

.github/workflows/tox-tests.yml
pllim: See https://github.com/astropy/specreduce/pull/202#discussion_r1418278247

pllim:
Sorry about the conflict with #182 . I don't watch this repo so I was not aware of it. I replied in https://github.com/astropy/specreduce/pull/202#discussion_r1418278247 (I think it is a little too soon but I am also not a specreduce maintainer).

Whether you want to merge this first, or cherry-pick your changes to this PR and take this over, I will leave to you.

specreduce/fluxcal.py
tepickering: i've done similar matplotlib tests in CI before and it's not that tricky (at least as far as checking whether the code works and makes a plot. testing if the plot is "correct" is a lot more complex...). definitely outside of the scope of this PR, though, so thanks for the issue.

tox.ini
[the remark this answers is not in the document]
tepickering: well, numpy has a much, much larger installed base to support and we don't. 3.10 isn't all that new so it's not a huge ask. if one cares at all about performance, they really want 3.11+, anyway.

tepickering:
no worries on the conflicts. i would say let's merge this ASAP and i'll do what i need to do to clean up #182. it's the changes that i've made in #182 that require 3.10+, iirc. if we want to stick with bumping the requirement there vs here, i'm fine with that. interested in @kecnry's take.

kecnry:
I don't have any strong opinions on the order of operations - good to merge on my end!  Thanks for all the improvements @pllim !

pllim:
Then, feel free to merge! I am tempted to merge myself but I shouldn't. Thanks!

PR #205, opened 2024-01-11
fix for masked values in FitTrace

cshanahan1:
This PR fixes some issues with FitTrace when bins are fully masked:

1. If `peak_method=gaussian`, the fit will fail if even one column/bin is fully masked, with the error being raised by the fitter. This PR fixes that issue by setting the fit trace value to nan for that column, so it can proceed and fit the next bin. For the final fit to all bin peaks for the trace, these nans bins/columns will then be filtered. 

2. If peak_method=max or centroid, a warning when a fully masked bin is encountered claiming that it is 'Falling back on trace value from all-bin fit.'. This is not accurate to what is happening
      2a. for `max`, the argmax of that bin will be returned which is always 0 for a fully masked bin. To avoid skewing the final fit to all bins when 0, i think a more appropriate value for this is 'nan' which can then be filtered from the final fit. theres nothing to really 'fall back' to in this case. This will require fixing almost all of the tests, so for now in this PR i just modify the warning message and retain the current behavior. See ISSUE.
       2b.  for `centroid`, when there is a fully masked bin it is also not 'falling back to the all bin fit'. to find the centroid of a fully masked bin, it attempts to interpolate between nans and will always return the maximum index in the bin. Like `max`, i am not changing the current behavior in this PR, that can be done as a follow up to resolve the corresponding ISSUE, but I change the warning message to reflect what is actually happening at the moment.

I also reorganized some stuff out of post_init into its own method, but this change is inconsequential.

### DIFF
--- a/specreduce/tracing.py
+++ b/specreduce/tracing.py
@@ -209,6 +209,8 @@ class FitTrace(Trace, _ImageParser):
     _disp_axis = 1
 
     def __post_init__(self):
+
+        # parse image
         self.image = self._parse_image(self.image)
 
         # mask any previously uncaught invalid values
@@ -262,14 +264,20 @@ class FitTrace(Trace, _ImageParser):
             warnings.warn('TRACE: Converting window to int')
             self.window = int(self.window)
 
+        # fit the trace
+        self._fit_trace(img)
+
+    def _fit_trace(self, img):
+
+        yy = np.arange(img.shape[self._crossdisp_axis])
+
         # set max peak location by user choice or wavelength with max avg flux
         ztot = img.sum(axis=self._disp_axis) / img.shape[self._disp_axis]
         peak_y = self.guess if self.guess is not None else ztot.argmax()
         # NOTE: peak finder can be bad if multiple objects are on slit
 
-        yy = np.arange(img.shape[self._crossdisp_axis])
-
         if self.peak_method == 'gaussian':
+
             # guess the peak width as the FWHM, roughly converted to gaussian sigma
             yy_above_half_max = np.sum(ztot > (ztot.max() / 2))
             width_guess = yy_above_half_max / gaussian_sigma_to_fwhm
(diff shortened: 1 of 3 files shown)


specreduce/tracing.py
```diff
@@ -302,15 +312,21 @@ def __post_init__(self):
 
         warn_bins = []
         for i in range(self.bins):
-            # repeat earlier steps to create gaussian fit for each bin
+
+            # binned column (sum) if bins < ncols. otherwise, just columns.
```
kecnry, line 316: can you elaborate more in this comment, this logic is hard to follow without really following where everything comes from above.

specreduce/tracing.py
```diff
@@ -336,24 +352,41 @@ def __post_init__(self):
...
+
             # if there are a ton of bins, we don't want to print them all out
             if len(warn_bins) > 20:
                 warn_bins = warn_bins[0: 10] + ['...'] + [warn_bins[-1]]
+
             warnings.warn(f"All pixels in {'bins' if len(warn_bins) else 'bin'} "
                           f"{', '.join([str(x) for x in warn_bins])}"
-                          " are masked. Falling back on trace value from all-bin fit.")
+                          " are fully masked. Setting bin"
+                          f" peak{'s' if len(warn_bins) else ''} to {warn_msg}.")
```
kecnry, line 389: The variable name `warn_msg` was confusing to me until I got all the way down here... can we use something more verbose/explanatory?  Maybe `warn_adopted_value` or something similar?  

Or maybe just do the formatting based on `peak_method` directly, with something like:


```
f" peak{'s' if len(warn_bins) else ''} to {{'gaussian': 'nan', 'max': 'zero'}.get(self.peak_method)}.")
```

kecnry (approved):

tepickering (approved):
thanks for this. looks good!

PR #209, opened 2024-02-08
Move some image parsing logic

cshanahan1:
Move part of the image parsing logic to its own function to avoid an issue in Background one_sided/two_sided.

Background.one_sided/two_sided were calling an instance method from _Image_Parser. This works as is, but if you were to then try to call another instance method from within that one, it does not understand 'self' and will cause an error. I moved the bit of code that Background needed to its own function to avoid this issue (which doesn't exist in main yet, but does in work on another branch I have).

Also fixed a bug where checking if data was `u.quantity.Quantity` was never triggered because the check for `np.ndarray` happened first.

I don't think this needs a change log.

### DIFF
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -9,7 +9,7 @@ from astropy.nddata import NDData
 from astropy.utils.decorators import deprecated_attribute
 from specutils import Spectrum1D
 
-from specreduce.core import _ImageParser
+from specreduce.core import _ImageParser, _get_data_from_image
 from specreduce.extract import _ap_weight_image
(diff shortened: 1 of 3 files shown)


cshanahan1:
> looks good! only comment would be to tweak the test to cover the missed line that codecov flagged.

i added some coverage and realized in the process that the `isinstance(image, u.quantity.Quantity)` was never actually being triggered because `isinstance(image, np.ndarray)` was being checked first. It didn't seem to cause any noticeable issues since this already existed on main, but I fixed it just in case.

PR #210, opened 2024-02-09
clean up image parsing and tests

cshanahan1:
1. Moved `_get_data_from_image`, which was previously part of _parse_image but used on its own in `Background`, to a method in _ImageParser instead of a function in the same module, to clean up imports.
2. Fixed some tests in `test_image_parsing`. Previously to test `_ImageParser.parse_image`, the method was being accessed through FlatTrace (which inherits _ImageParser and therefore has parse_image), and 'object' was used seemingly to avoid failing when 'self' was accessed. This was working for the time being, but will cause issues when other methods are called within _parse_image (like the one added in this PR)

### DIFF
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -9,7 +9,7 @@ from astropy.nddata import NDData
 from astropy.utils.decorators import deprecated_attribute
 from specutils import Spectrum1D
 
-from specreduce.core import _ImageParser, _get_data_from_image
+from specreduce.core import _ImageParser
 from specreduce.extract import _ap_weight_image
 from specreduce.tracing import Trace, FlatTrace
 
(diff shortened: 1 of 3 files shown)


tepickering (approved):
looks good and cleans up the namespace. one question: if a class has a leading _, do its static and other methods need them as well? i would lean towards "no", but don't feel that strongly about it.

cshanahan1:
> if a class has a leading _, do its static and other methods need them as well? i would lean towards "no", but don't feel that strongly about it.

thats a good question, i can see the case for both ways. if the class is 'private' i guess it would be implied its methods are too, so another leading underscore would be redundant? but it also would seem logical for them to have a leading underscore as well. in another PR i can do some cleanup and make sure everything follows one convention or the other.

PR #211, opened 2024-02-20
fix input for background.traces, raise error in FlatTrace for negative trace

cshanahan1:
This PR cleans up a few things in Background.

1. Added a condition to FlatTrace that trace position must be negative. This check was originally done in Background when a FlatTrace was being used, but should generally be done. Before, if trace_pos was negative, a fully masked trace was being returned.

2. Previously, Background produced an unrelated error if 'traces' was None. Now, if no trace/number is passed in, a FlatTrace in the middle of the image will be used. 

3. There was some other inconsistencies with inputs - the docstring said 'traces' must be a list, but the logic in the code implied that it could be a Trace object not in a list. Also, a FlatTrace object can be defined with either a float or an int, where previously 'Background' allowed only integer inputs. I fixed this to allow either single trace/int/float or list trace/int/float inputs. The only condition is that lists must all be of the same type (either numeric or `Trace`). I also added some checks for input to make sure something valid is passed in for `traces` (either None, int/float or list of int/float to define FlatTraces, or a Trace object, or a list of Trace objects). I also added tests for all of these allowed inputs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,9 @@ Bug Fixes
   were set to the number of rows in the image, biasing the final fit to all bin
   peaks. Previously for Gaussian, the entire fit failed. [#205, #206]
 
+- Fixed input of `traces` in `Background`. Added a condition to 'FlatTrace' that
+  trace position must be a positive number. [#211]
+
 Other changes
 ^^^^^^^^^^^^^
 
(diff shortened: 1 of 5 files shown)


tepickering (changes requested):
the changes look good to me. thanks for the cleanup and fixes. however, something about the changes caused the RTD build to fail. i looked at the raw build logs and found the warnings that point to `tracing.py`, but don't quite understand why it's not happy.

cshanahan1:
fixed the failing docstring

PR #214, opened 2024-03-13
Add a function to measure cross-dispersion profile

cshanahan1:
This PR adds a function to measure the cross-dispersion profile at a single wavelength, or the average profile in a set/range of wavelengths. It takes in a 'width' for the window in the cross dispersion axis, which moves around the trace. A 1D array is returned with either the profile at that single wavelength/pixel or the mean or median profile across several wavelength/pixels.

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -118,34 +118,6 @@ def _ap_weight_image(trace, width, disp_axis, crossdisp_axis, image_shape):
     return wimage
 
 
-def _align_along_trace(img, trace_array, disp_axis=1, crossdisp_axis=0):
-    """
-    Given an arbitrary trace ``trace_array`` (an np.ndarray), roll
-    all columns of ``nddata`` to shift the NDData's pixels nearest
-    to the trace to the center of the spatial dimension of the
-    NDData.
-    """
-    # TODO: this workflow does not support extraction for >2D spectra
-    if not (disp_axis == 1 and crossdisp_axis == 0):
-        # take the transpose to ensure the rows are the cross-disp axis:
-        img = img.T
-
-    n_rows, n_cols = img.shape
-
-    # indices of all columns, in their original order
-    rows = np.broadcast_to(np.arange(n_rows)[:, None], img.shape)
-    cols = np.broadcast_to(np.arange(n_cols), img.shape)
-
-    # we want to "roll" each column so that the trace sits in
-    # the central row of the final image
-    shifts = trace_array.astype(int) - n_rows // 2
-
-    # we wrap the indices so we don't index out of bounds
-    shifted_rows = np.mod(rows + shifts[None, :], n_rows)
-
-    return img[shifted_rows, cols]
-
-
 @dataclass
 class BoxcarExtract(SpecreduceOperation):
     """
@@ -814,6 +786,34 @@ class HorneExtract(SpecreduceOperation):
         # convert the extraction to a Spectrum1D object
         return Spectrum1D(extraction * unit,
                           spectral_axis=self.image.spectral_axis)
+
+
+def _align_along_trace(img, trace_array, disp_axis=1, crossdisp_axis=0):
+    """
+    Given an arbitrary trace ``trace_array`` (an np.ndarray), roll
+    all columns of ``nddata`` to shift the NDData's pixels nearest
+    to the trace to the center of the spatial dimension of the
+    NDData.
+    """
+    # TODO: this workflow does not support extraction for >2D spectra
+    if not (disp_axis == 1 and crossdisp_axis == 0):
+        # take the transpose to ensure the rows are the cross-disp axis:
+        img = img.T
+
+    n_rows, n_cols = img.shape
+
+    # indices of all columns, in their original order
+    rows = np.broadcast_to(np.arange(n_rows)[:, None], img.shape)
+    cols = np.broadcast_to(np.arange(n_cols), img.shape)
+
+    # we want to "roll" each column so that the trace sits in
+    # the central row of the final image
+    shifts = trace_array.astype(int) - n_rows // 2
+
+    # we wrap the indices so we don't index out of bounds
+    shifted_rows = np.mod(rows + shifts[None, :], n_rows)
+
+    return img[shifted_rows, cols]
 
 
 @dataclass
--- a/specreduce/utils/__init__.py
+++ b/specreduce/utils/__init__.py
@@ -1,3 +1,5 @@
 """
 General purpose utilities for specreduce
 """
+
+from .utils import *  # noqa
--- a/specreduce/utils/utils.py
+++ b/specreduce/utils/utils.py
@@ -1,0 +1,212 @@
+import numpy as np
+
+from specreduce.core import _ImageParser
+from specreduce.tracing import Trace, FlatTrace
+from specreduce.extract import _ap_weight_image, _align_along_trace
+
+__all__ = ['measure_cross_dispersion_profile', '_align_along_trace']
+
+
+def measure_cross_dispersion_profile(image, trace=None, crossdisp_axis=0,
+                                     width=None, pixel=None, pixel_range=None,
+                                     statistic='median', align_along_trace=True):
+    """
+    Return a 1D (quantity) array of the cross-dispersion profile measured at a
+    specified pixel value ('wavelength', but potentially before calibration),
+    or the average profile across several pixel values (or range of pixel values)
+    along the dispersion axis.
+
+    If a single number is specified for `pixel`, then the profile at that pixel
+    (i.e wavelength) will be returned. If several pixels are specified in a list or
+    array, then they will be averaged (median or mean, set by `statistic` which
+    defaults to median). Alternatively, `pixel_range` can be specified as a tuple
+    of integers specifying the minimum and maximum pixel range to average the
+    profile. `pixel` and `pixel_range` cannot be set simultaneously, and the default
+    is `pixel_range`=(min_image, max_image) to return an average profile across the
+    entire wavelength range of the image.
+
+    The window in the cross dispersion axis for measuring the profile at
+    the pixel(s) specified is determined by `width` and `trace`. If a trace is
+    provided (either as a Trace object, or as a number specifying the location
+    on the cross-dispersion axis of a FlatTrace object) that will determine the
+    center of the profile on the cross-dispersion axis. Otherwise, if `trace`
+    is None, the center of the image will be the center of the returned profile
+    (i.e., the center row assuming a horizontal trace).
+
+    If `width` is none, a window size of half the image in the cross-dispersion
+    axis will be used to measure the cross dispersion profile. Otherwise, an
+    integer value can be set for `width` which will determine the size of the
+    window around the trace used to measure the profile (this window moves with
+    the trace if trace is not flat).
+
+    By default, if a non-flat trace is used the image will be aligned along the
+    trace. This can be controlled with the 'align_along_trace' parameter.
+
+    Parameters
+    ----------
+    image : `~astropy.nddata.NDData`-like or array-like, required
+        2D image to measure cross-dispersion profile.
+    trace : Trace object, int, float, or None
+        A Trace object, a number to specify a FlatTrace, or None to use the
+        middle of the image. This is the position that defines the center of the
+        cross-dispersion profile. [default: None]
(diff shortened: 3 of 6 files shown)


specreduce/utils/utils.py
```diff
@@ -0,0 +1,170 @@
...
+                                     width=None, pixel=None, pixel_range=None,
+                                     statistic='median'):
+    """
+    Return a 1D (quantity) array of the cross-dispersion profile measured at a
+    specified pixel value ('wavelength', but potentially before calibration),
+    or the average profile across several pixel values (or range of pixel values)
+    along the dispersion axis.
+
+    If a single number is specified for `pixel`, then the profile at that pixel
+    (i.e wavelength) will be returned. If several pixels are specified in a list or
```
camipacifici, line 19: If pixels are mapped to wavelength, can user input the wavelength directly? Does it need units?

specreduce/utils/utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+    If `width` is none, a window size of half the image in the cross-dispersion
+    axis will be used to measure the cross dispersion profile. Otherwise, an
+    integer value can be set for `width` which will determine the size of the
+    window around the trace used to measure the profile (this window moves with
+    the trace if trace is not flat).
+
+    Parameters
+    ----------
+    image : `~astropy.nddata.NDData`-like or array-like, required
```
camipacifici, line 43: Trace and Background accept a Spectrum1D object too. Can this work in the same way? If I try feeding a Spectrum1D object, I get this traceback:

MaskError                                 Traceback (most recent call last)
Cell In[19], line 1
----> 1 profile = cdprofile(objectspec, trace)

File [~/opt/miniconda3/envs/prsr214/lib/python3.11/site-packages/specreduce/utils/utils.py:157](http://localhost:8888/~/opt/miniconda3/envs/prsr214/lib/python3.11/site-packages/specreduce/utils/utils.py#line=156), in measure_cross_dispersion_profile(image, trace, crossdisp_axis, width, pixel, pixel_range, statistic)
    154 # combine these masks to isolate the rows and cols used to measure profile
    155 combined_mask = np.logical_or(pixel_mask, wimg)
--> 157 masked_arr = np.ma.MaskedArray(image, combined_mask)
    159 # and measure the cross dispersion profile. if multiple pixels[/wavelengths](http://localhost:8888/wavelengths),
    160 # this will be an average. we already transposed data based on disp_axis so
    161 # axis is always 1 for this calculation
    162 if statistic == 'mean':

File [~/opt/miniconda3/envs/prsr214/lib/python3.11/site-packages/numpy/ma/core.py:2911](http://localhost:8888/~/opt/miniconda3/envs/prsr214/lib/python3.11/site-packages/numpy/ma/core.py#line=2910), in MaskedArray.__new__(cls, data, mask, dtype, copy, subok, ndmin, fill_value, keep_mask, hard_mask, shrink, order)
   2908     else:
   2909         msg = "Mask and data not compatible: data size is %i, " + \
   2910               "mask size is %i."
-> 2911         raise MaskError(msg % (nd, nm))
   2912     copy = True
   2913 # Set the mask to the new value

MaskError: Mask and data not compatible: data size is 1, mask size is 2000.

cshanahan1: Yeah I am working on this - I need to move what is in _parse_image somewhere else so this can use it

specreduce/utils/utils.py
```diff
@@ -0,0 +1,181 @@
...
+import numpy as np
+
+from specreduce.core import _ImageParser
+from specreduce.tracing import Trace, FlatTrace
+from specreduce.extract import _ap_weight_image
+
+
+def measure_cross_dispersion_profile(image, trace=None, crossdisp_axis=0,
+                                     width=None, pixel=None, pixel_range=None,
+                                     statistic='median'):
```
kecnry, line 12: is this intended to be public immediately or are we keeping it internal for now?  Should it be added to `__all__` and the docs if it is public?

specreduce/utils/utils.py
```diff
@@ -0,0 +1,181 @@
...
+    of integers specifying the minimum and maximum pixel range to average the
+    profile. `pixel` and `pixel_range` cannot be set simultaneously, and the default
+    is `pixel_range`=(min_image, max_image) to return an average profile across the
+    entire wavelength range of the image.
+
+    The window in the cross dispersion axis for measuring the profile at
+    the pixel(s) specified is determined by `width` and `trace`. If a trace is
+    provided (either as a Trace object, or as a number specifying the location
+    on the cross-dispersion axis of a FlatTrace object) that will determine the
+    center of the profile on the cross-dispersion axis. Otherwise, if `trace`
```
kecnry, line 32: is this done by shifting pixels or does it support subpixel via interpolation, etc?  Or - unless this logic is within some other function that I forget about - is this just using the trace to create the mask but then not doing any recentering/shifting at all?

specreduce/utils/utils.py
```diff
@@ -0,0 +1,181 @@
...
+        is also None, this will be interpreted as using the entire dispersion axis
+        to generate an average profile. [default: None]
+    statistic: 'median' or 'mean'
+        If `pixel` specifies multiple pixels, or `pixel_range` is specified, an
+        average profile will be returned. This can be either `median` (default)
+        or `mean`. This is ignored if only one pixel is specified. [default: median]
+    """
+
+    # work on copy of image so input isn't modified
+    image = copy.copy(image)
```
kecnry, line 73: would it be modified otherwise?  Or does either `_parse_image` or `MaskedArray` create a copy anyways?

specreduce/utils/utils.py
```diff
@@ -0,0 +1,181 @@
...
+                             'the location of a FlatTrace, or None to use center'
+                             ' of image.')
+
+    if statistic not in ['median', 'mean']:
+        raise ValueError("`statistic` must be 'median' or 'mean'.")
+
+    # determine if there is one pixel/wavelength selected or many as either a
+    # list or a tuple to specify a range
+    if pixel is not None:
+        if not (pixel is None) ^ (pixel_range is None):
```
kecnry, line 116: can this second line be simplified given that it's within `pixel is not None`?

kecnry (commented):
Do we want this as a method on any of the extraction steps, perhaps on trace itself?  If there are other uses, it makes sense to have the general capability as a standalone function, but it might be convenient to be able to call `trace.cross_dispersion_profile` and/or `trace.plot_cross_dispersion_profile()`, etc.

specreduce/utils/utils.py
cshanahan1: i don't see why it can't be public, ill add it. 

in terms of making this part of the extraction steps, i think it can still live in 'utils' as a standalone and we can add methods to the steps to call it if that would be helpful. should that be done in this ticket or a follow up?

specreduce/utils/utils.py
cshanahan1: yeah i think so

specreduce/utils/utils.py
cshanahan1: This is just using the trace to create the mask

kecnry: Is that sufficient?  I thought I remember hearing the request for subpixel support along the trace, but can't remember if that was deferred or not.

kecnry: To follow-up: the plan for now will be to implement an option for aligning by the trace by pixel-shifting (integer, not subpixel).

cshanahan1: i added this functionality with the 'align_along_trace' arg

cshanahan1:
@camipacifici @kecnry this is ready for a final review - i added some of the tests for a non-flat trace with and without aligning the image along the trace.

the only outstanding item is the pixels vs wavelengths thing. right now, since its clear that the input is pixels, i personally think its OK to add wavelength as another parameter (that can't be set simultaneously with pixels) later on. what do you think?

cshanahan1:
test data with a non-flat trace:
Screen Shot 2024-03-25 at 10 19 12 AM

the returned profile with align_along_trace = False, at different pixels. the center moves as expected:
Screen Shot 2024-03-25 at 10 19 53 AM

And with align_along_trace = True - since the trace is rolled to the nearest integer value, the center for the profile is the same + or - one pixel:
Screen Shot 2024-03-25 at 10 19 43 AM

specreduce/extract.py
```diff
@@ -118,34 +118,6 @@ def _ap_weight_image(trace, width, disp_axis, crossdisp_axis, image_shape):
...
-    # we want to "roll" each column so that the trace sits in
-    # the central row of the final image
-    shifts = trace_array.astype(int) - n_rows // 2
-
-    # we wrap the indices so we don't index out of bounds
-    shifted_rows = np.mod(rows + shifts[None, :], n_rows)
-
-    return img[shifted_rows, cols]
-
-
```
kecnry, line 148: was this just moved in the file or were there changes made here (hard to tell from the diff)?

specreduce/utils/utils.py
```diff
@@ -0,0 +1,209 @@
...
+from specreduce.core import _ImageParser
+from specreduce.tracing import Trace, FlatTrace
+from specreduce.extract import _ap_weight_image, _align_along_trace
+
+__all__ = ['measure_cross_dispersion_profile', '_align_along_trace']
+
+
+def measure_cross_dispersion_profile(image, trace=None, crossdisp_axis=0,
+                                     width=None, pixel=None, pixel_range=None,
+                                     statistic='median', align_along_trace=False):
```
kecnry, line 12: Sounds like the consensus is to change the default to `align_along_trace=True` (API docstring also needs updating)
```suggestion
                                     statistic='median', align_along_trace=True):
```

specreduce/utils/utils.py
```diff
@@ -0,0 +1,209 @@
...
+            trace = FlatTrace(aligned_trace, trace_pos)
+
+    # create a weight image based on the trace and 'width' to mask around trace
+
+    if width == nrows:
+        wimg = np.zeros(image.shape)
+    else:
+        wimg = _ap_weight_image(trace, width, disp_axis, crossdisp_axis, image.shape)
+        # invert mask to include, not exclude, pixels around trace
+        wimg = (1 - wimg).astype(int)
```
kecnry, line 179: would it be hard to get coverage for this scenario?

specreduce/extract.py
cshanahan1: just moved it. i had it in utils but there was a circular import so i moved it back.

specreduce/utils/utils.py
cshanahan1: Done

specreduce/utils/utils.py
cshanahan1: done

kecnry (approved):
LGTM now, thanks!

camipacifici:
Looks great, thank you!

camipacifici (approved):

specreduce/utils/__init__.py
```diff
@@ -1,3 +1,5 @@
 """
 General purpose utilities for specreduce
 """
+
+from .utils import *  # noqa
```
tepickering, line 5: do we want relative imports here?

specreduce/utils/utils.py
tepickering: my take is that the core routine for this should work in pixel space. if we want to work in wavelength space, then we can wrap or decorate this to accept a WCS and input wavelengths to determine `pixel` and `pixel_range`.

tepickering (approved):
good stuff! thanks for this. a couple minor comments, but otherwise is good to go as far as i'm concerned.

specreduce/utils/utils.py
cshanahan1: are you ok with that being a follow up effort?

specreduce/utils/__init__.py
cshanahan1: i'm not sure?

kecnry: I don't have an opinion either... we can always change this later if we come up with a policy/standard we want to follow, right?

cshanahan1:
merging this - wavelength input can be addressed in a follow up

specreduce/utils/__init__.py
tepickering: well, the PEP8 guide recommends absolute imports, fwiw:  https://peps.python.org/pep-0008/#imports.

PR #215, opened 2024-03-27
update namespace of Background in the documentation (docs/extraction_quickstart.rst)

rudolffu:
Replace `specreduce.tracing.Background` with `specreduce.background.Background` for correct calling of the `Background` submodule.


kecnry (approved):
Good catch - thanks!

PR #216, opened 2024-04-22
Add default and set to zero masking option for specreduce operations.

cshanahan1:
Adds the ability to specify how to handle boolean masks to background, trace, and extract operations. 'filter', 'omit' to drop entire columns that contain any nonfinite data, or zero-fill to replace nonfinite data with zeros.


tepickering:
finally got a chance to at least look through the code changes. i like it so far! let me know when it's ready for a fuller review and i'll try playing with it in more depth.

camipacifici:
I am pretty sure it is! Please go ahead. Thank you!

tepickering:
there are a bunch of warnings in the RTD build. they look like malformed references, but need a deeper look to see what's the actual cause.

tepickering:
i'm confused why python 3.8 and 3.9 tests are being run. a rebase might be needed, though i thought this was forked after those tests were removed.

specreduce/background.py
```diff
@@ -32,9 +32,9 @@ class Background(_ImageParser):
     ----------
     image : `~astropy.nddata.NDData`-like or array-like
         image with 2-D spectral image data
-    traces : trace, int, float (single or list)
+    traces : List, `tracing.Trace`, int, float
```
tepickering, line 35: i think this is one of the RTD complaints. should be `~specreduce.tracing.Trace`, i believe.

specreduce/background.py
```diff
@@ -32,9 +32,9 @@ class Background(_ImageParser):
...
         Individual or list of trace object(s) (or integers/floats to define
-        FlatTraces) to extract the background. If None, a FlatTrace at the
+        FlatTraces) to extract the background. If None, a `FlatTrace` at the
```
tepickering, line 37: `should be ``FlatTrace`` or a proper sphinx reference`

specreduce/background.py
```diff
@@ -44,8 +44,22 @@ class Background(_ImageParser):
         pixels.
     disp_axis : int
         dispersion axis
+        [default: 1]
     crossdisp_axis : int
         cross-dispersion axis
+        [default: 0]
+    mask_treatment : string, optional
+        The method for handling masked or non-finite data. Choice of `filter`,
```
tepickering, line 52: here and elsewhere need to use double single-quotes for inline literal/code. pretty sure this is what RTD has been complaining about...

specreduce/core.py
```diff
@@ -67,14 +68,20 @@ def _get_data_from_image(image, disp_axis=1):
...
+        mask = getattr(image, 'mask', None)
+
+        # next, handle masked and nonfinite data in image.
+        # A mask will be created from any nonfinite image data, and combined
+        # with any additional 'mask' passed in. If image is being parsed within
+        # a specreduce operation that has 'mask_treatment' options, this will be
+        # handled as well. Note that input data may be modified if a fill value
+        # is chosen to handle masked data. The returned image will always have
+        # `image.mask` even if there are no nonfinte or masked values.
+        img, mask = self._mask_and_nonfinite_data_handling(image=img, mask=mask)
```
hpparvi, line 80: `_get_data_from_image` is a static method, so `self` is not defined. This could be fixed by making `_mask_and_nonfinite_data_handling` a static method as well and calling it as `_ImageParser._mask_and_nonfinite_data_handling(...)`, but this would require removing the `self.image.mask` reference in line 144. 

Also, `_ImageParser` never assigns anything to `self.image`. This could be fixed by changing the line 55

    img = self._get_data_from_image(image, disp_axis=disp_axis)
    return img

into

    self.image = self._get_data_from_image(image, disp_axis=disp_axis)
    return self.image

and declaring self.image in the  `_ImageParser` initializer.

specreduce/core.py
```diff
@@ -85,11 +92,106 @@ def _get_data_from_image(image, disp_axis=1):
...
 
-        return Spectrum1D(img * unit, spectral_axis=spectral_axis,
-                          uncertainty=uncertainty, mask=mask)
+        img = Spectrum1D(img * unit, spectral_axis=spectral_axis,
+                         uncertainty=uncertainty, mask=mask)
+
+        return img
+
+    @staticmethod
+    def _get_data_from_image(image):
```
hpparvi, line 101: This method definition overrides the previous `_get_data_from_image` definition that handles masking.

specreduce/core.py
cshanahan1: This has become kind of a mess and i'm trying to figure out how to reorganize. A while back I had to move `_get_data_from_image` to a static method because some of the 'Background' methods that create Background objects call it before the object is instantiated. I am trying to work off the existing code and avoid refactoring in this PR, but I may have to. This option I'm exploring right now is changing _ImageParser from a mixin class to either just a regular class or set of functions that is called within each step, and the arguments are passed in each time.

hpparvi: The only way to avoid refactoring `_ImageParser` that I can see would be to turn `_mask_and_nonfinite_data_handling` into a static method as well and to provide `mask_treatment` as a passed argument (and to handle checking whether the mask treatment method is valid in the calling code.) This way, the `_get_data_from_image` class method could be used the way it is now. 

However, there wouldn't be much advantage to having the image parsing functionality as a class instead of a function, and I think refactoring `_ImageParser` into a function might make the most sense. Turning it into a regular class would also work, but I'm not sure what would be again the advantage of this over having the image parsing as a function if we want to be able to parse an image inside a class method (as with the `Background` methods). 

Then again, we don't necessarily need to do the image parsing in the `Background.two_sided` and `Background.one_sided` class methods since it is done in `Background.__post_init__` by default (unless I'm missing something). So, another option would be to leave the image parsing inside `Background.__post__init__` and remove it from the two class methods, in which case the `Background` object would always be instantiated before parsing the image.

If you want to avoid major refactors in this PR, I can create a separate PR where I refactor the `_ImageParser` to a function, and we can see how much sense this approach would make. What do you think?

hpparvi (approved):

hpparvi:
Great work, @cshanahan1, I'm merging this now.

PR #221, opened 2024-06-17
fix issue with units

cshanahan1:
fixes a few things

- background one sided / two sided not properly assigning units to data when new image is passed in.
- fixed check for quantity vs array in parse data
- add change log section for 1.4.1


tepickering (approved):
nice catch. looks good and ready for version bump.

PR #224, opened 2024-08-12
TST: Use photutils nightly wheel

pllim:
https://github.com/astropy/photutils/pull/1834


tepickering (approved):
thanks for this! not sure why `codecov` is complaining about this PR given the only change is in `tox.ini`. i'd say we ignore that and merge anyway.

pllim:
Yeah, I don't quite trust codecov project coverage anymore. There was a whole conversation on Astropy to replace it but I don't think there is a replacement that would do everything this codecov is currently doing yet (like the visual inspection stuff). See #16619

pllim:
Thanks for review!

PR #228, opened 2024-11-11
Image data handling refactor

hpparvi:
This draft PR proposes one way to refactor image data handling in Specreduce. The PR introduces a `specreduce.image.SRImage` class that works as a (relatively thin) wrapper around `astropy.nddata.NDDataRef`. I chose to wrap `NDDataRef` instead of `specutils.Spectrum1D` because I think this is a more appropriate class for 2D images, especially when we start working on 2D wavelength calibration. The methods that produce 1D spectra still return `Spectrum1D` objects.

The main functionalities of `SRImage` are:
- The `SRImage` initialiser combines the image parsing logic that was previously in `specreduce.core._ImageParser` and `specreduce.extract.HorneExtract._parse_image`.
- The initialiser standardises the image axes so that the cross-dispersion axis is always the first axis and the dispersion axis is always the second axis. The primary motivation for this is to ensure the rest of the code doesn't need to worry about the ordering of the axes. Also, the re-ordering is done using `numpy.moveaxis` instead of `numpy.transpose` in the case we want to support images with `ndim > 2` in the future. (The class has `crossdisp_axis` and `disp_axis` properties, which are fixed to 0 and 1, respectively.)
- The class contains an `apply_mask` method that applies the mask to the image data using the logic introduced by @cshanahan1 in PR #216, and a `to_masked_array`method that applies the mask and returns the data as a masked array.

The PR modifies the `Background`, `Tracing`, and `Extraction` classes to work with the `SRImage` class. The refactor passes all the tests without modifications to the tests themselves (well,  except the ones for `HorneExtract`, which is something I still need to work on), so the changes should not modify or break the functionality of any existing code using public methods.

While it passes (nearly) all the tests, **this PR is not ready to be merged.** I've created it to raise discussion and hear everyone's opinion about the approach to the refactor. So, please comment on whether this is a sensible way to go or if you'd prefer something else.

If this approach gains sufficient support, I'll polish the code and prepare the PR ready to be merged. In any case, this will need to wait until PR #216 is merged (the code is based on the masking work by @cshanahan1 in #216), and I still need to document the code properly and rewrite some of the `HorneExtract` tests to work with the new image handling logic, but my aim is that the `HorneExtraction` class itself will work identically as before without any visible changes to the end-user.


pllim:
The commit history of this PR is dirty. Please rebase.

Issue #227, opened 2024-11-11
python 3.11 CI test failure


kecnry:
`specreduce/tests/test_synth_data.py::test_make_2d_arc_pass_wcs` is now failing with:

```
astropy.units.core.UnitConversionError: Can only apply 'greater' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan).
```

during a call to specutils

```
../../specreduce/utils/synth_data.py:290: in make_2d_arc_image
    linelist = load_pypeit_calibration_lines(linelists, wave_air=wave_air)
../../specreduce/calibration_data.py:288: in load_pypeit_calibration_lines
    linelist['wavelength'] = vac_to_air(linelist['wavelength'])
../../.tox/py311-test-alldeps-cov/lib/python3.11/site-packages/specutils/utils/wcs_utils.py:148: in vac_to_air
    refr = refraction_index(wavelength, method=method, co2=co2)
../../.tox/py311-test-alldeps-cov/lib/python3.11/site-packages/specutils/utils/wcs_utils.py:103: in refraction_index
    if method != 'Griesen2006' and wavelength.min() < 200 * u.nm:
```

pllim:
Is this `linelist['wavelength']` not Quantity?

hpparvi:
This is now fixed in PR #231.

PR #231, opened 2024-11-15
Fix for PypeIt calibration line loading

hpparvi:
This PR fixes Issue #227 by converting the `linelist` table in `specreduce.calibration_data.load_pypeit_calibration_lines` from a Table to a QTable before calling `specutils.utils.wcs_utils.vac_to_air`. The `vac_to_air` function requires wavelengths to be given as `Quantity` objects, but the line list was a Table instead of a QTable.


kecnry (approved):
Thanks for the quick fix!

PR #232, opened 2024-11-15
Fix for Astropy v7.0 incompatibility (#229)

### LINKED ISSUE #229: astropy v7.0 incompatibility in TestMasksTracing.test_fit_trace_all_nan_cols
According to [astropy-integration-testing run](https://github.com/astropy/astropy-integration-testing/actions/runs/11785052451/job/32825493170), there is failure with both stable and dev versions, so this package needs a compatibility fix and a new release to be compatible with `astropy` v7.0.0. FYI.

hpparvi:
This PR fixes Issue #229 by increasing the absolute and relative tolerances of `assert_allclose` in `test_fit_trace_all_nan_col`. An absolute tolerance of 0.05 pixels should be sufficient to test that the tracing methods work correctly but leave enough room for any minor variations in the results caused by the changes in the LevMarLSQFitter between Astropy v6 and v7.

Fix #229

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,24 +10,32 @@ API Changes
 Bug Fixes
 ^^^^^^^^^
 
+Other changes
+^^^^^^^^^^^^^
+
+1.4.2
+------
+
+Bug Fixes
+^^^^^^^^^
+- Fixed Astropy v7.0 incompatibility bug [#229] in ``tracing.FitTrace``: changed to use
+  ``astropy.modeling.fitting.DogBoxLSQFitter`` when fitting a Gaussian peak model instead of
+  ``astropy.modeling.fitting.LevMarLSQFitter`` that may be deprecated in the future. Also
+  changed to use ``fitting.LMLSQFitter`` instead of ``fitting.LevMarLSQFitter`` when fitting
+  a generic nonlinear trace model.
 
 Other changes
+^^^^^^^^^^^^^
+- Changed ``tracing.FitTrace`` to use ``astropy.modeling.fitting.LinearLSQFitter``
+  if the trace model is linear.
 
 1.4.1 (2024-06-20)
 ------------------
-
-New Features
-^^^^^^^^^^^^
-
-API Changes
-^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


pllim:
Please rebase to pick up the other CI fix.

hpparvi:
> The broader question, though, is whether specreduce should still be using a fitter that is no longer recommended by core astropy library ([astropy/astropy#16983](https://github.com/astropy/astropy/pull/16983)).

It probably shouldn't. Fixing this should be relatively trivial, and I can do it either in this PR or a separate one.

pllim:
Actually I am wondering also if relaxing tolerance is even needed if you switch fitter class.

hpparvi:
It seems still to be required. I've changed the fitter and run the tests, and the results still differ slightly (at a 0.01-pixel level).

pllim:
I'll defer to specreduce maintainers on which fitters to replace the old one with. Thanks!

hpparvi:
Any opinions, @cshanahan1, @kecnry, @tepickering? I think we can adopt `DogBoxLSQFitter` without any problems.

hpparvi:
> agreed this is top priority to fix ASAP. i'd go ahead and make the switch to `DogBoxLSQFitter`.

Thanks @tepickering! I've changed the fitting method and also added a check to use `LinearLSQFitter` for linear trace models like Chebyshev1D.

@pllim, this PR should be ready to merge now. I haven't been granted a maintainer status quite yet, but the two-week comment period is over now without any opposing comments (https://github.com/astropy/astropy.github.com/pull/639), so this should change soon.

pllim (approved):
Thanks!

tepickering (approved):
thanks for catching all this! i like adding the checks to help choose the optimal fitter.

tepickering:
> Does this need a change log?

i think so since it's a pretty significant change to how model fitting is done. once that's added, i can go ahead and merge.

pllim:
Since Tim said he would merge, I'll give him the honors. But if he is MIA, please ping me again because I can also merge but would rather let package maintainers do this. Thanks, all!

Issue #169, opened 2023-04-06
Refactor reference file loading to use pure astropy


tepickering:
The current implementation of `calibration_data.get_reference_file_path` wraps `astropy.utils.data.download_file` with bespoke error handling and URL management. We should be able to use the higher level methods in `astropy.utils.data` instead.

PR #165, opened 2023-03-13
Synthetic wavelength calibration data

tepickering:
This PR adds a utility for creating realistic models of 2D spectroscopic images of arc lamp (or sky background) calibration observations. This will be key for testing the various stages of the wavelength calibration code and process as it's developed. Tilts/curvature along the spatial axis is modeled using 1D polynomials. Non-linear dispersion relations are modeled using the ideal disperser model implemented within the FITS WCS standard. Examples of the use of this routine are included in its docstrings. 

A few other notes:

- Use of the `specreduce_data` dependency has been deprecated. Cleaner to just always download/cache the data as needed. 
- https://github.com/pypeit/PypeIt is used as the source for comparison line lists because it has a large selection of well-vetted ones. 
- In the example where a non-linear dispersion relation is created using the WCS ideal disperser model, I create the model by putting together the appropriate FITS header manually and then loading that into `astropy.wcs.WCS`. It would be useful to have a documented functional interface for building WCS spectrograph models that provides meaningful parameter names and units handling. However, I think that is outside of the scope of this PR. It's also an open question whether such a utility should live here or within `specutils`.

PR #233, opened 2024-11-18
Refactor reference file loading

hpparvi:
This PR addresses Issue #169 (also discussed in PR #165) by using the `astropy.config.ConfigNamespace.set_temp` context manager to temporarily set the default data download URL and then using either `astropy.utils.data.get_pkg_data_fileobj` or `astropy.utils.data.get_pkg_data_filename` to get a readable file or a path to the downloaded and cached file.

However, I'm not convinced this approach leads to more elegant code than the original one. We eliminate the two wrapper functions, but the file-handling code is somewhat messy because of all the nesting. Please take a look and let me know what you think.


hpparvi:
Do you have any opinions about this, @cshanahan1, @tepickering, and @kecnry? If nobody opposes this approach, I'll improve the code coverage and see if I can make the code a bit more readable.

tepickering:
thanks for catching this! i filed the original issue, but had long since forgotten about it 😅. i like what you've done so far. will give it a closer look when you're more ready to merge. a few test failures to address as well as the coverage.

hpparvi:
Thanks, @tepickering. However, the more I looked at the approach, the less I liked it. The code's readability decreased too much without any improvements in the functionality. I've now switched back to using download_file consistently but without the wrappers (and added a couple of new tests checking that we catch URLErrors from bad file names correctly), which leads to cleaner and more readable code. 

The PR is ready to merge in theory, but the code doesn't differ much from the original. It mainly just skips the use of the download wrapper functions. If you still think this is worth merging, doing a squash merge is probably best because most of the intermediate commits were very small.

tepickering (approved):
looks good. my only comment of note is that some of the diffs are due to changes in code style. it may be worth following what `astropy` does and use `black` and/or `ruff` to define the style and apply it automatically and consistently.

tepickering:
> Thanks, @tepickering. However, the more I looked at the approach, the less I liked it. The code's readability decreased too much without any improvements in the functionality. I've now switched back to using download_file consistently but without the wrappers (and added a couple of new tests checking that we catch URLErrors from bad file names correctly), which leads to cleaner and more readable code.
> 
> The PR is ready to merge in theory, but the code doesn't differ much from the original. It mainly just skips the use of the download wrapper functions. If you still think this is worth merging, doing a squash merge is probably best because most of the intermediate commits were very small.

it might, in fact, be worthwhile to make squash merging the default going forward. right now all three methods are allowed and i can't see how to change the default to "squash". i think it becomes the default if i disable "merge commit", but not sure if we want to pull that trigger yet. i'm fine with keeping our options open and deciding on a per-PR basis, but i'm open to other ideas.

PR #245, opened 2025-01-29
Boxcar extraction with non-finite pixels

hpparvi:
This PR adds a new `exclude` mask treatment option to BoxcarExtract, allowing the extraction to exclude any non-finite pixels. With the `exclude` option, the extraction is now carried out as a weighted sum, calculated as the average of the finite in-window pixels multiplied by the number of in-window pixels. The behaviour stays identical to the previous with the other mask treatment options or if no non-finite values are present.

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -149,15 +149,15 @@ class BoxcarExtract(SpecreduceOperation):
         cross-dispersion axis
     mask_treatment
         The method for handling masked or non-finite data. Choice of ``filter``,
-        ``omit``, or ``zero-fill``. If `filter` is chosen, the mask is ignored
+        ``omit``, or ``exclude``. If ``filter`` is chosen, the mask is ignored
         and the non-finite data will passed to the extraction as is. If ``omit``
         is chosen, columns along disp_axis with any masked or non-finite data
         values will be fully masked (i.e, 2D mask is collapsed to 1D and applied).
-        If ``zero-fill`` is chosen, masked and non-finite data will be replaced
-        with 0.0 in the input image, and the mask will then be dropped.
-        For all three options, the input mask (optional on input NDData object)
-        will be combined with a mask generated from any non-finite values in the
-        image data.
+        If ``exclude`` is chosen, masked and non-finite data will be excluded
+        from the extraction and the spectrum extraction is carried out as a
+        weighted sum. For all three options, the input mask (optional on input
+        NDData object) will be combined with a mask generated from any non-finite
+        values in the image data.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


hpparvi:
I improved the documentation slightly to describe how the non-finite values can be handled in boxcar extraction.

specreduce/extract.py
```diff
@@ -170,8 +170,9 @@ class BoxcarExtract(SpecreduceOperation):
     disp_axis: int = 1
     crossdisp_axis: int = 0
     # TODO: should disp_axis and crossdisp_axis be defined in the Trace object?
-    mask_treatment: Literal['filter', 'omit', 'zero-fill'] = 'filter'
-    _valid_mask_treatment_methods = ('filter', 'omit', 'zero-fill')
+    # TODO: should the 'filter' option be changed to 'propagate'
+    mask_treatment: Literal['filter', 'omit', 'exclude'] = 'exclude'
```
tepickering, line 174: this choice of words doesn't make it immediately clear what each option does.

tepickering (changes requested):
i find that the names of the mask treatment options as implemented are not very clear and in fact somewhat misleading. it's worth having longer names that are more clearly descriptive of what they actually do. and if only two options are functionally supported, there should only be two options available. in my opinion, the options that make sense are:

- `ignore`: don't apply mask or mask non-finite values; extract from image as-is.
- `apply`: apply mask and mask non-finite values as you've implemented (should be default).
- `apply_mask_only`: apply mask, but ignore non-finite values
- `mask_nan_only`: ignore mask, but mask non-finite values
- `mask_extraction_bin`: grow mask so that if any masked or non-finite values are in the extraction bin, the whole bin gets masked in the output.

`apply_mask_only` and `mask_nan_only` are kind of niche cases so maybe we don't worry about those. `zero_fill` is another option that might make sense in some contexts, but i think masking properly is better. 

there's also the larger, more contentious issue that `BoxcarExtract` is a class that really should be a function. all of its functionality is contained within `__call__()` and making it a class adds a redundant layer of type declarations. it also requires users to either use two lines of code to perform an extraction or an unintuitive second pair of `()`'s.

camipacifici:
This discussion makes sense and here is my proposal at matching your suggestions:
- `filter` can be renamed with `ignore`
- `apply` should be default and it is the new behavior (right?)
- I would keep the `zero-fill` option with its name since it is clear
- I like the idea of including `apply_mask_only` and `mask_nan_only`
- I am not totally sold on `mask_extraction_bin`...the behavior would be like `apply` but for the whole bin, so how about `apply_to_full_extraction_bin` instead? It is kind of long though. This would be the same as `omit` right?
Thanks!

kecnry:
I like @camipacifici's suggestions for option naming.

> If there are good reasons to keep them as classes (anyone?)

This is related to the decision of whether we want to support reusable interactive modifications of inputs with internal caching, or not.  If we decide "not", then I think functions would be reasonable and likely would simplify the code itself 🤷‍♂️

hpparvi:
Great, I also like @camipacifici's suggestions. I've opened a new Issue (#252) and will submit a new PR with revised code and documentation in a few days.

@tepickering, would you be okay with merging the current PR while we address the mask treatment options in a separate PR? I’d prefer to refactor the entire codebase and documentation systematically.

tepickering:
works for me! go for it.

hpparvi:
Thanks! Can you still approve the PR for merging?

tepickering (approved):
looks good for now. we can address naming and other concerns in future PR as discussed.

Issue #249, opened 2025-02-10
Horne extraction skips columns with any non-finite values


hpparvi:
Currently, Horne extraction skips columns with non-finite values. A better approach would be to exclude these values from the extraction and renormalize the extraction weights, as is already done in boxcar extraction. This should be rather trivial to fix.

PR #250, opened 2025-02-12
Issue 249 fix

hpparvi:
This PR addresses Issue #249 to make Horne extraction ignore non-finite values in the data instead of propagating them into the spectrum.

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -605,70 +605,35 @@ class HorneExtract(SpecreduceOperation):
         disp_axis = disp_axis if disp_axis is not None else self.disp_axis
         crossdisp_axis = crossdisp_axis if crossdisp_axis is not None else self.crossdisp_axis
         bkgrd_prof = bkgrd_prof if bkgrd_prof is not None else self.bkgrd_prof
-        spatial_profile = (spatial_profile if spatial_profile is not None else
-                           self.spatial_profile)
+        profile = (spatial_profile if spatial_profile is not None else self.spatial_profile)
         variance = variance if variance is not None else self.variance
         mask = mask if mask is not None else self.mask
         unit = unit if unit is not None else self.unit
 
-        # figure out what 'spatial_profile' was provided
-        # put this parsing into another method at some point, its a lot..
-        interp_degree_interpolated_profile = None
-        n_bins_interpolated_profile = None
-        spatial_profile_choices = ('gaussian', 'interpolated_profile')
+        profile_choices = ('gaussian', 'interpolated_profile')
 
-        if isinstance(spatial_profile, str):
-            spatial_profile = spatial_profile.lower()
-            if spatial_profile not in spatial_profile_choices:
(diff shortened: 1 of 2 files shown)


specreduce/extract.py
```diff
@@ -605,70 +610,35 @@ def __call__(self, image=None, trace_object=None,
...
+        profile_choices = ('gaussian', 'interpolated_profile')
+
+        if not isinstance(profile, (str, dict)):
             raise ValueError('``spatial_profile`` must either be string or dictionary.')
+        if isinstance(profile, str):
+            profile = dict(name=profile)
+
+        profile_type = profile['name'].lower()
+        if profile_type not in profile_choices:
+            raise ValueError("spatial_profile must be one of" f"{', '.join(profile_choices)}")
```
tepickering, line 627: want to add a test case to cover this case...

specreduce/extract.py
```diff
@@ -735,75 +694,42 @@ def __call__(self, image=None, trace_object=None,
...
-                    fit_ext_kernel.mean_0 = mean_init_guess[col_pix]
-                else:
-                    fit_ext_kernel.mean = mean_init_guess[col_pix]
+        if profile_type == 'gaussian':
+            norms[:] = fit_ext_kernel.amplitude_0 * fit_ext_kernel.stddev_0 * np.sqrt(2*np.pi)
 
-                # evaluate fit model (with shifted mean, based on trace)
+        for icol in range(ncols):
+            if not np.any(valid[:, icol]):
+                continue
```
tepickering, line 717: add test case to cover the codecov complaint

specreduce/extract.py
```diff
@@ -482,53 +482,58 @@ def _fit_gaussian_spatial_profile(self, img: ndarray, disp_axis: int, crossdisp_
...
+        -------
+        RectBivariateSpline
+            Interpolator object that provides a smoothed 2D spatial profile.
+        """
 
-        # need to make a new masked array since this mask is different
-        # omit all columns with nans from contributing to the fit
-        new_masked_arr = np.ma.array(img.data.copy(), mask=img_col_mask)
+        img = np.where(~mask, img, np.nan)
+        nrows = img.shape[crossdisp_axis]
```
tepickering, line 524: probably better to use something like `ncross` and `ndisp` because dispersion is not always aligned with rows...

tepickering (approved):
a few minor changes (missing coverage; fix assumption of row-aligned dispersion in variable names), but otherwise looks good. punting on extraction bins that had any masked/non-finite values was a cop-out in most cases so this fix is much appreciated.

specreduce/extract.py
hpparvi: Added.

specreduce/extract.py
hpparvi: Added.

specreduce/extract.py
hpparvi: True, I've changed the variable names.

PR #253, opened 2025-02-27
implement average and median support for bkg_spectrum

gibsongreen:
This pull request is created to add functionality to the bkg_spectrum function. Currently, this function utilizes np.nansum for all computations of the background spectrum. This pull request defaults this function with np.nansum, while also adding support for 'Average' and 'Median' by utilizing np.nanmean and np.nanmedian.

### DIFF
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -365,7 +365,8 @@ class Background(_ImageParser):
         elif bkg_statistic == 'average':
             statistic_function = np.nanmean
         else:
-            raise ValueError(f"Background statistics not supported for {bkg_statistic}")
+            raise ValueError(f"Background statistic {bkg_statistic} is not supported. "
+                             "Please choose from: average, median, or sum.")
 
         try:
             return bkg_image.collapse(statistic_function, axis=self.crossdisp_axis)
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -342,11 +342,12 @@ class Background(_ImageParser):
             direction is axis 1. If None, will extract the background
             from ``image`` used to initialize the class. [default: None]
         bkg_statistic : str, optional
-            Statistical method used to collapse the background image.
+            Statistical method used to collapse the background image. [default: ``sum``]
             Supported values are:
-            - `'median'` : Uses the median (`np.nanmedian`).
-            - `'average'` : Uses the mean (`np.nanmean`).
-            - Any other value (including `None`): Defaults to summation (`np.nansum`).
+
+            - `'median'` : Uses the median (`numpy.nanmedian`).
+            - `'average'` : Uses the mean (`numpy.nanmean`).
+            - `'sum'` : Uses the sum (`numpy.nansum`).
 
         Returns
         -------
@@ -360,7 +361,6 @@ class Background(_ImageParser):
         statistic_function = np.nansum
 
         if bkg_statistic:
-            bkg_statistic = bkg_statistic.lower()
             if bkg_statistic == 'median':
                 statistic_function = np.nanmedian
             elif bkg_statistic == 'average':
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -345,9 +345,9 @@ class Background(_ImageParser):
             Statistical method used to collapse the background image. [default: ``sum``]
             Supported values are:
 
-            - `'median'` : Uses the median (`numpy.nanmedian`).
-            - `'average'` : Uses the mean (`numpy.nanmean`).
-            - `'sum'` : Uses the sum (`numpy.nansum`).
+            - ``median`` : Uses the median (`numpy.nanmedian`).
+            - ``average`` : Uses the mean (`numpy.nanmean`).
+            - ``sum`` : Uses the sum (`numpy.nansum`).
 
         Returns
         -------
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -330,7 +330,7 @@ class Background(_ImageParser):
                                   (image.shape[0], 1)) * image.unit,
                           spectral_axis=image.spectral_axis)
 
-    def bkg_spectrum(self, image=None, bkg_statistic=None):
+    def bkg_spectrum(self, image=None, bkg_statistic="sum"):
         """
         Expose the 1D spectrum of the background.
 
@@ -341,13 +341,13 @@ class Background(_ImageParser):
             (spatial) direction is axis 0 and dispersion (wavelength)
             direction is axis 1. If None, will extract the background
             from ``image`` used to initialize the class. [default: None]
-        bkg_statistic : str, optional
-            Statistical method used to collapse the background image. [default: ``sum``]
+        bkg_statistic : {'average', 'median', 'sum'}, optional
+            Statistical method used to collapse the background image. [default: ``'sum'``]
             Supported values are:
 
-            - ``median`` : Uses the median (`numpy.nanmedian`).
-            - ``average`` : Uses the mean (`numpy.nanmean`).
-            - ``sum`` : Uses the sum (`numpy.nansum`).
+            - ``'average'`` : Uses the mean (`numpy.nanmean`).
+            - ``'median'`` : Uses the median (`numpy.nanmedian`).
+            - ``'sum'`` : Uses the sum (`numpy.nansum`).
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


specreduce/background.py
```diff
@@ -341,6 +341,12 @@ def bkg_spectrum(self, image=None):
             (spatial) direction is axis 0 and dispersion (wavelength)
             direction is axis 1. If None, will extract the background
             from ``image`` used to initialize the class. [default: None]
+        bkg_statistic : str, optional
+            Statistical method used to collapse the background image.
+            Supported values are:
+            - `'median'` : Uses the median (`np.nanmedian`).
```
pllim, line 347: Ref linking needs full name of the package. Please fix accordingly to get rid of warnings on ReadTheDocs. Thanks!
```suggestion
            - `'median'` : Uses the median (`numpy.nanmedian`).
```

specreduce/background.py
```diff
@@ -341,6 +341,12 @@ def bkg_spectrum(self, image=None):
...
+            - `'average'` : Uses the mean (`np.nanmean`).
+            - Any other value (including `None`): Defaults to summation (`np.nansum`).
```
pllim, line 349: Why not just say `'sum'` here and use that as default? What is the benefit of allowing `None` or any other value here?

specreduce/background.py
pllim: I also wonder if passing in the function directly is more straightforward than using string for translation. It certainly would be more flexible and not confined to numpy. User can pass in their custom stats func... but is that too dangerous?

specreduce/background.py
```diff
@@ -351,12 +357,21 @@ def bkg_spectrum(self, image=None):
         """
         bkg_image = self.bkg_image(image)
 
+        statistic_function = np.nansum
+
+        if bkg_statistic:
+            bkg_statistic = bkg_statistic.lower()
```
pllim, line 363: I don't think we need to support mixed case here. I say we only support inputs in all lowercase. Python is naturally case-sensitive.

```suggestion
```

specreduce/background.py
gibsongreen: I was thinking about the implementation in Jdaviz (passing the traitlet value for Statistic) but can see the applicability to be able to pass the function directly too. Could modify the jdaviz logic to make this possible so I'm open to either way.

specreduce/background.py
gibsongreen: I used `lower` here for the Jdaviz traitlet which has the first letter capitalized in the string, that could be done in Jdaviz instead, so I'm good with removing this and moving the logic elsewhere (or like you said passing the function directly)

specreduce/background.py
gibsongreen: The default is sum, I was thinking that if the user passed a value that wasn't in the available options that it would still return a Spectrum1D object, I thought about throwing an exception if that were the case but was trying to avoid it.

specreduce/background.py
pllim: I don't think the API here should be dictated by what Jdaviz does. The API we pick should be useful to someone who is using just specreduce natively. Jdaviz can hook into it downstream no matter what.

gibsongreen: I've made the other changes locally, I'll leave the string in for now but am more than open to modifying the logic here to make it more flexible

pllim: Let's see what specreduce maintainers say. Thanks!

specreduce/background.py
```diff
@@ -330,7 +330,7 @@ def bkg_image(self, image=None):
                                   (image.shape[0], 1)) * image.unit,
                           spectral_axis=image.spectral_axis)
 
-    def bkg_spectrum(self, image=None):
+    def bkg_spectrum(self, image=None, bkg_statistic=None):
```
pllim, line 333:
```suggestion
    def bkg_spectrum(self, image=None, bkg_statistic="sum"):
```

specreduce/background.py
```diff
@@ -341,6 +341,13 @@ def bkg_spectrum(self, image=None):
             (spatial) direction is axis 0 and dispersion (wavelength)
             direction is axis 1. If None, will extract the background
             from ``image`` used to initialize the class. [default: None]
+        bkg_statistic : str, optional
+            Statistical method used to collapse the background image. [default: ``sum``]
+            Supported values are:
+
+            - `'median'` : Uses the median (`numpy.nanmedian`).
+            - `'average'` : Uses the mean (`numpy.nanmean`).
+            - `'sum'` : Uses the sum (`numpy.nansum`).
```
pllim: Alphabetical?

```suggestion
            - ``'average'`` : Uses the mean (`numpy.nanmean`).
            - ``'median'`` : Uses the median (`numpy.nanmedian`).
            - ``'sum'`` : Uses the sum (`numpy.nansum`).
```

specreduce/background.py
```diff
@@ -341,6 +341,13 @@ def bkg_spectrum(self, image=None):
             (spatial) direction is axis 0 and dispersion (wavelength)
             direction is axis 1. If None, will extract the background
             from ``image`` used to initialize the class. [default: None]
+        bkg_statistic : str, optional
+            Statistical method used to collapse the background image. [default: ``sum``]
```
pllim, line 345:
```suggestion
        bkg_statistic : {'average', 'median', 'sum'}, optional
            Statistical method used to collapse the background image. [default: ``'sum'``]
```

specreduce/background.py
```diff
@@ -351,12 +358,20 @@ def bkg_spectrum(self, image=None):
         """
         bkg_image = self.bkg_image(image)
 
+        statistic_function = np.nansum
+
+        if bkg_statistic:
+            if bkg_statistic == 'median':
+                statistic_function = np.nanmedian
+            elif bkg_statistic == 'average':
+                statistic_function = np.nanmean
```
pllim, line 367:
```suggestion
        if bkg_statistic == 'sum':
            statistic_function = np.nansum
        elif bkg_statistic == 'median':
            statistic_function = np.nanmedian
        elif bkg_statistic == 'average':
            statistic_function = np.nanmean
        else:
            raise ValueError(f"Background statistics not supported for {bkg_statistic}")
```

specreduce/tests/test_background.py
```diff
@@ -80,6 +81,12 @@ def test_background(mk_test_img_raw, mk_test_spec_no_spectral_axis,
         assert np.isnan(bg.bkg_spectrum().flux).sum() == 0
         assert np.isnan(bg.sub_spectrum().flux).sum() == 0
 
+    bkg_spec_avg = bg1.bkg_spectrum(bkg_statistic='Average')
```
pllim, line 84:
```suggestion
    bkg_spec_avg = bg1.bkg_spectrum(bkg_statistic='average')
```

specreduce/tests/test_background.py
```diff
@@ -80,6 +81,12 @@ def test_background(mk_test_img_raw, mk_test_spec_no_spectral_axis,
...
+    assert_allclose(bkg_spec_avg.mean().value, 14.5, rtol=0.5)
+
+    bkg_spec_median = bg1.bkg_spectrum(bkg_statistic='Median')
```
pllim, line 87:
```suggestion
    bkg_spec_median = bg1.bkg_spectrum(bkg_statistic='median')
```

pllim (commented):
Might also want to add a test for invalid input and ensure ValueError is raised. Thanks!

gibsongreen:
> Might also want to add a test for invalid input and ensure ValueError is raised. Thanks!

Will do, thank you for the thorough review, it was very helpful!

cshanahan1 (approved):
This looks good to me, thanks! I added a couple of minor suggestions.

specreduce/background.py
gibsongreen: @cshanahan1 I had that originally but removed it with this is mind

specreduce/background.py
tepickering: i like the idea of passing in a function for full generality, but also recognize that in this context, mean and median are the most applicable options in the the vast majority of cases.

tepickering (approved):
thanks for this. i am approving this, but want to note that "average" and "median" return a different scale than "sum". i.e., they're normalized to the area of the background region and thus easily scaled to a source region for subtraction. "sum" is not scaled and the number of background pixels used to calculate it would be required to use it for background subtraction. tracking number of pixels excluded by `np.nansum` is tricky and possibly complicated. i'm open to keeping "sum" as an option, but i'm unclear on the use case.

gibsongreen:
> thanks for this. i am approving this, but want to note that "average" and "median" return a different scale than "sum". i.e., they're normalized to the area of the background region and thus easily scaled to a source region for subtraction. "sum" is not scaled and the number of background pixels used to calculate it would be required to use it for background subtraction. tracking number of pixels excluded by `np.nansum` is tricky and possibly complicated. i'm open to keeping "sum" as an option, but i'm unclear on the use case.

Thank you for reviewing! This function solely used sum prior to this PR, which was my reasoning in keeping it in and leaving it as the default parameter. I had considered removing sum altogether, however, there is test coverage for sum that results in several test failures and I didn't want to break existing functionality. 

I'm open to making any additional changes necessary in this PR or a follow-up effort, whether that is setting the default parameter for the statistical function to be either mean/median or remove the option for sum.

tepickering:
> I'm open to making any additional changes necessary in this PR or a follow-up effort, whether that is setting the default parameter for the statistical function to be either mean/median or remove the option for sum.

yes, i think it's a larger issue for discussion so it's fine to leave it for now and do a follow-up effort later. i would want to revisit why "sum" was used in the first place before nixing it.

Issue #252, opened 2025-02-21
Change mask treatment option names


hpparvi:
The mask treatment option names should be changed to `apply`, `ignore`, `zero-fill`, `apply_mask_only`, `apply_nan_only`, and `apply_to_full_exctraction_bin`, as discussed in PR #245.

PR #254, opened 2025-02-27
Mask and non-finite option refactor

hpparvi:
This PR addresses Issue #252 by renaming the existing mask treatment options and introducing a new set. It also adds a documentation page with illustrations demonstrating each option's effect in practice, and it also adds a basic Black configuration to `pyproject.toml` and modifies flake8 in `tox.ini` to ignore E203, as described [here](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#labels-why-pycodestyle-warnings). Finally, I started running black on the code I edited, which leads to somewhat larger diffs than from the actual functional changes alone.

The complete mask treatment option set now includes: `apply`, `ignore`, `propagate`, `zero-fill`, `nan-fill`, `apply_mask_only`, and `appy_nan_only`. Now is the best time to let everyone know if you are not happy with some of these choices, so please write if you have strong feelings against any :)

I chose `propagate` over the suggested `mask_extraction_bin` and `apply_to_full_extraction_bin` because it is more generic and can be used in any method or function requiring this behavior, rather than being limited to spectrum extraction. However, I'm happy to change this to something else if anyone can suggest a good, clear alternative.

Finally, should we go for `apply_mask_only` or `apply-mask-only`? The first option is probably grammatically better, but the other would be easier to remember.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,11 +21,14 @@ New Features
 
 - Added the ``mask_treatment`` parameter to Background, Trace, and Boxcar Extract
   operations to handle non-finite data and boolean masks. Available options are
-  ``filter``, ``omit``, or ``zero-fill``, with ``exclude`` additionally available
-  for BoxcarExtract. [#216]
+  ``apply``, ``ignore``, ``propagate``, ``zero-fill``, ``nan-fill``, ``apply_mask_only``,
+  or ``apply_nan_only``. [#216, #254]
 
-- Modified BoxcarExtract to ignore non-finite pixels when ``mask_treatment`` is set
-  to ``exclude``; otherwise, non-finite values are propagated. Boxcar extraction is
+- Modified ``background.Background.bgk_spectrum`` to allow the user to select the statistic
+  used for background estimation between ``median`` or ``average``. [#253]
+
+- Modified ``extract.BoxcarExtract`` to ignore non-finite pixels when ``mask_treatment`` is set
+  to ``apply``; otherwise, non-finite values are propagated. Boxcar extraction is
   now carried out as a weighed sum over the window. When no non-finite values are
   present, the extracted spectra remain unchanged from the previous behaviour.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,7 +21,7 @@ New Features
 
 - Added the ``mask_treatment`` parameter to Background, Trace, and Boxcar Extract
   operations to handle non-finite data and boolean masks. Available options are
-  ``apply``, ``ignore``, ``propagate``, ``zero-fill``, ``nan-fill``, ``apply_mask_only``,
+  ``apply``, ``ignore``, ``propagate``, ``zero_fill``, ``nan_fill``, ``apply_mask_only``,
   or ``apply_nan_only``. [#216, #254]
 
 - Modified ``background.Background.bgk_spectrum`` to allow the user to select the statistic
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ build-backend = 'setuptools.build_meta'
 
 [tool.black]
 line-length = 100
-target-version = ['py310', 'py311', 'py312']
+target-version = ['py311', 'py312', 'py313']
 
 [tool.pytest.ini_options]
 minversion = 7.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,10 @@ requires = ["setuptools",
 ]
 build-backend = 'setuptools.build_meta'
 
+[tool.black]
+line-length = 100
+target-version = ['py310', 'py311', 'py312']
+
 [tool.pytest.ini_options]
 minversion = 7.0
 testpaths = [
--- a/specreduce/core.py
+++ b/specreduce/core.py
@@ -181,12 +181,12 @@ class _ImageParser:
         """
         if mask_treatment not in _ImageParser.implemented_mask_treatment_methods:
             raise ValueError(
-                "`mask_treatment` must be one of "
+                "'mask_treatment' must be one of "
                 f"{_ImageParser.implemented_mask_treatment_methods}"
             )
 
         if mask is not None and (mask.dtype not in (bool, int)):
-            raise ValueError("`mask` must be a boolean or integer array.")
+            raise ValueError("'mask' must be a boolean or integer array.")
 
         match mask_treatment:
             case "apply":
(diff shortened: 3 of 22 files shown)


hpparvi:
Looks like I still need to fix some minor issues with the docs. Here is a print of the doc page illustrating the mask treatment options.

[mask_treatment.pdf](https://github.com/user-attachments/files/19015619/mask_treatment.pdf)

hpparvi:
Also, I still need to improve the tests for the new code.

tepickering:
> Also, I still need to improve the tests for the new code.

shall i wait for you to do this before doing a review?

hpparvi:
Okay, the PR is now ready for review, @tepickering .

CHANGES.rst
```diff
@@ -21,11 +21,14 @@ New Features
 
 - Added the ``mask_treatment`` parameter to Background, Trace, and Boxcar Extract
   operations to handle non-finite data and boolean masks. Available options are
-  ``filter``, ``omit``, or ``zero-fill``, with ``exclude`` additionally available
-  for BoxcarExtract. [#216]
+  ``apply``, ``ignore``, ``propagate``, ``zero-fill``, ``nan-fill``, ``apply_mask_only``,
```
tepickering, line 24: i think we should pick ``_`` or ``-`` and use that consistently. i did some quick searching and it looks like ``numpy`` and ``matplotlib`` use ``_``, though ``scipy`` uses both. i have a soft preference for ``_`` since it looks more pythonic and is often more shell-friendly.

pyproject.toml
```diff
@@ -57,6 +57,10 @@ requires = ["setuptools",
 ]
 build-backend = 'setuptools.build_meta'
 
+[tool.black]
+line-length = 100
+target-version = ['py310', 'py311', 'py312']
```
tepickering, line 62: should add ``py313`` here as well and i think it's safe to drop ``py310``.

specreduce/core.py
```diff
@@ -247,9 +244,11 @@ def as_function(cls, *args, **kwargs):
         """
         argspec = inspect.getargs(SpecreduceOperation.__call__.__code__)
         if argspec.varargs:
-            raise NotImplementedError('There is not a way to determine the '
-                                      'number of inputs of a *args style '
-                                      'operation')
+            raise NotImplementedError(
```
tepickering, line 247: also needs test.

specreduce/extract.py
```diff
@@ -611,25 +646,25 @@ def __call__(self, image=None, trace_object=None,
...
         variance = variance if variance is not None else self.variance
         mask = mask if mask is not None else self.mask
         unit = unit if unit is not None else self.unit
 
-        profile_choices = ('gaussian', 'interpolated_profile')
+        profile_choices = ("gaussian", "interpolated_profile")
 
         if not isinstance(profile, (str, dict)):
-            raise ValueError('``spatial_profile`` must either be string or dictionary.')
+            raise ValueError("``spatial_profile`` must either be string or dictionary.")
```
tepickering, line 657: add test

specreduce/extract.py
```diff
@@ -645,29 +680,27 @@ def __call__(self, image=None, trace_object=None,
...
+                mean_cross_pix = np.broadcast_to(ncross // 2, ndisp)
         else:  # interpolated_profile
             # for now, bkgrd_prof must be None because a compound model can't
             # be created with a interpolator + model. i think there is a way
             # around this, but will follow up later
             if bkgrd_prof is not None:
-                raise ValueError('When `spatial_profile`is `interpolated_profile`,'
-                                 '`bkgrd_prof` must be None. Background should'
-                                 ' be fit and subtracted from `img` beforehand.')
+                raise ValueError(
```
tepickering, line 699: add test

specreduce/tests/test_extract.py
```diff
@@ -89,19 +89,15 @@ def test_boxcar_nonfinite_handling(mk_test_img):
...
     target = np.full_like(spectrum.flux.value, 90.)
     target[2] = np.nan
     target[4] = np.inf
     np.testing.assert_equal(spectrum.flux.value, target)
 
-    trace = FlatTrace(image, 15.0)
-    boxcar = BoxcarExtract(image, trace, width=6, mask_treatment='exclude')
-    spectrum = boxcar()
-    target = np.full_like(spectrum.flux.value, 90.)
+    spectrum = BoxcarExtract(image, trace, width=6, mask_treatment='apply')()
```
tepickering, line 98: was ``black`` run on this file? i would have thought it would change to double-quotes here.

tepickering (approved):
i'm happy with these changes and the choice of option names. just need to pick ``_`` or ``-``. i also really appreciate setting up ``black`` so that we can keep our codestyle consistent. i left some comments where coverage is missing on some exception cases. i would recommend setting up test(s) to cover those now or they'll get forgotten.

CHANGES.rst
hpparvi: I also prefer `_` and have now changed `nan-fill` to `nan_fill` and `zero-fill` to `zero_fill`.

specreduce/tests/test_extract.py
hpparvi: No, it wasn't. This is fixed now.

specreduce/extract.py
hpparvi: Added!

specreduce/extract.py
hpparvi: Added!

pyproject.toml
hpparvi: Done.

specreduce/core.py
```diff
@@ -233,9 +230,9 @@ class SpecreduceOperation(_ImageParser):
     the operation, which then *return* the data objects resulting from the
     operation.
     """
+
     def __call__(self):
-        raise NotImplementedError('__call__ on a SpecreduceOperation needs to '
-                                  'be overridden')
+        raise NotImplementedError("__call__ on a SpecreduceOperation needs to " "be overridden")
```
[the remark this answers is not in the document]
hpparvi: Added.

specreduce/core.py
hpparvi: Added.

specreduce/extract.py
```diff
@@ -392,37 +411,41 @@ def _parse_image(self, image,
...
+        if hasattr(image, "uncertainty") and image.uncertainty is not None:
+            if image.uncertainty.uncertainty_type == "var":
                 variance = image.uncertainty.array
-            elif image.uncertainty.uncertainty_type == 'std':
-                warnings.warn("image NDData object's uncertainty "
-                              "interpreted as standard deviation. if "
-                              "incorrect, use VarianceUncertainty when "
-                              "assigning image object's uncertainty.")
+            elif image.uncertainty.uncertainty_type == "std":
+                warnings.warn(
```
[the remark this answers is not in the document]
hpparvi: Added.

pllim:
I think this broke Jdaviz -- https://github.com/spacetelescope/jdaviz/pull/3487

hpparvi:
Thanks for the heads up, @pllim. This should now be fixed in PR #256.

PR #255, opened 2025-02-28
bkg_spectrum to take func instead of string

pllim:
This is a direct follow-up of unreleased feature added in https://github.com/astropy/specreduce/pull/253 so behavior change can be done without deprecation or change log.

I am just wondering what it would have looked like.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-1.5.1 (unreleased)
+1.5.2 (unreleased)
 ------------------
 
 New Features
@@ -7,8 +7,13 @@ New Features
 API Changes
 ^^^^^^^^^^^
 
+- ``bkg_statistic`` keyword in ``Background.bkg_spectrum()`` now takes callable
+  instead of string. Its default is also changed from ``"sum"`` (``np.nansum``)
+  to ``np.nanmedian`` to better suit background calculation. [#253]
+
 Bug Fixes
 ^^^^^^^^^
+
 - When all-zero bin encountered in fit_trace with peak_method=gaussian, the bin peak
   will be set to NaN in this caseto work better with DogBoxLSQFitter. [#257]
 
--- a/specreduce/background.py
+++ b/specreduce/background.py
@@ -311,7 +311,7 @@ class Background(_ImageParser):
             spectral_axis=image.spectral_axis,
         )
 
-    def bkg_spectrum(self, image=None, bkg_statistic=np.nansum):
+    def bkg_spectrum(self, image=None, bkg_statistic=np.nanmedian):
         """
         Expose the 1D spectrum of the background.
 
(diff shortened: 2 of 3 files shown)


kecnry:
If we do this, I think we'd want to mirror that in the class-level `statistic` parameter, but that has additional logic and restriction to just median/average, which might be difficult to make fully general 🤷‍♂️

pllim:
Where is the `statistic`? I was looking at `Spectrum1D.collapse` upstream which takes function.

kecnry:
In `Background` (see https://github.com/astropy/specreduce/blob/main/specreduce/background.py#L41, for example)

tepickering (changes requested):
there is a complaint about a conflict that needs resolving before we can merge. i would suggest making `np.nanmedian` the default vs `np.nansum`. we want a per-pixel background in most cases and the median statistic is more robust against outliers that can skew the mean. personally, i would use `scipy.stats.mode` in many cases and i appreciate how this change would allow me to do so.

pllim:
@tepickering , I think I have addressed your comment. But now though, this is a breaking change. Not sure if it belongs in bugfix anymore... 😬

hpparvi:
Hi, I'm afraid there might be some confusion regarding the logic and intended functionality of the `bgk_spectrum` method (sorry, I should've caught this already with #253). 

- Background estimation occurs in `__post_init__` using the `statistic` parameter, which currently accepts either `average` or `median`. The resulting 1D background estimate is stored in `Background._bkg_array`.

- The `Background.bkg_image` method creates a 2D background image by tiling the 1D array (`Background._bkg_array`) along the cross-dispersion axis. This results in a 2D `Spectrum1D` object whose pixel values vary along the dispersion axis but remain constant along the cross-dispersion axis.

- The `Background.bkg_spectrum` method collapses the 2D background image from `Background.bkg_image` back into a 1D `Spectrum1D` by applying a statistical function along the cross-dispersion axis. However, since all pixels along the cross-dispersion axis have identical values, choosing different statistical functions here has no practical effect.

If we want to change the statistic used for background estimation, we need to do it in `Background.__post_init__`, as @pllim pointed out earlier. We could change `statistic` to accept either a string or a custom function that accepts masked arrays without affecting existing functionality.

 @kecnry, you introduced bkg_spectrum in c53ea53d3a58c54cb6d576b8d5b99f1a74a4de60. Can you remind us of the original use-case for the method?

kecnry:
Yes, I definitely think with the past few PRs all being somewhat related that we need to make sure we have a plan for what this all means and how it fits together.  I'll do some digging with @gibsongreen @cshanahan1 and some old notes and will get back to you then.

hpparvi:
Here is my follow-up continuing @pllim's work. I've rebased the branch onto the current ``main`` with compatibility updates for specutils v2, modified the ``statistic`` parameter in the ``Background`` initializer to accept a callable, updated the docstrings, and adjusted ``Background.bkg_spectrum`` to issue a warning if ``bkg_statistic`` is set. 

https://github.com/hpparvi/specreduce/tree/pllim-hpparvi-flexi-bgfunc

@pllim, would you prefer that I open a PR against your branch (this can't be merged automatically), or should I create a new PR directly to ``astropy/specreduce:main``?

pllim:
New PR is fine. I can close this one. Thanks for taking this over!

PR #256, opened 2025-03-08
Horne extraction fix

hpparvi:
This PR reverts Horne extraction to its earlier behaviour for cases when an interpolated spatial profile is used and the background profile isn’t explicitly set to None. The revised behaviour had led to [issues in jdaviz](https://github.com/spacetelescope/jdaviz/pull/3487)  without any added advantages.

### DIFF
--- a/specreduce/extract.py
+++ b/specreduce/extract.py
@@ -299,9 +299,10 @@ class HorneExtract(SpecreduceOperation):
         The index of the image's cross-dispersion axis. [default: 0]
 
     bkgrd_prof : `~astropy.modeling.Model` or None, optional
-        A model for the image's background flux. If ``spatial_profile`` is set
-        to ``interpolated_profile``, then ``bkgrd_prof`` must be set to None.
-        [default: models.Polynomial1D(2)].
+        A model for the image's background flux when using the ``gaussian``
+        spatial profile. If ``spatial_profile`` is set to ``gaussian``, it defaults
+        to ``models.Polynomial1D(2)``. Note that the ``interpolated_profile`` option
+        does not support a background model, so ``bkgrd_prof`` must be left as ``None``.
 
     spatial_profile : str or dict, optional
         The shape of the object profile. The first option is 'gaussian' to fit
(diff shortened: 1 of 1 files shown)


hpparvi:
This is a small fix that stops specreduce raising an error when using an interpolated profile but not explicitly setting the background profile to None in Horne extraction (the way it was before v1.5.0). This should fix the jdaviz issue. I'll create a specreduce release v1.5.1 after this PR has been approved.

pllim:
Thank you! Jdaviz should test this patch downstream before merging this here. Pinging @cshanahan1 because she was going to pick the work up.

kecnry:
We have max-pinned specreduce and are plan to update to handle the latest release next week... so I'm not sure that reverting is necessary on our end.  @cshanahan1 - what are your thoughts?

pllim:
This looks like only a small revert. Is it that undesirable?

cshanahan1 (commented):
Would it be better to check if the model is not a polynomial with coefficients all 0, and still keep this error to catch cases where a user does try to set a background model? Or what about the default actually being None rather than Polynomial(000)?

pllim:
@hpparvi , are you able to temporarily copy over the latest tag from this repo to your fork? I am enable to directly test this branch over at https://github.com/spacetelescope/jdaviz/pull/3492 because setuptools_scm unable to find tag on your fork, so it thinks the specreduce version is 0.1 and that failed glue-astronomy pin. Thanks for your help!

hpparvi:
> Would it be better to check if the model is not a polynomial with coefficients all 0 rather than None (since it will never actually be None), and still keep this error to catch cases where a user does try to set a background model? Or what about the default actually being None rather than Polynomial(000)?

Good point @cshanahan1. The last option makes the most sense to me, so I've now changed the background model initialisation default to `None`. If the spatial profile is `gaussian` and a background model isn't given explicitly, the background model is now set to `Polynomial1D(2)` in `__call__`. This doesn't change the code's behaviour but is a bit more elegant than the current approach. I've also changed the documentation to explain the default behaviour.

hpparvi:
@pllim, can you take a look now?

cshanahan1 (approved):
this fixes the failing test in jdaviz, thank you!

pllim:
I trust Clare. :shipit:  Thanks!

PR #262, opened 2025-04-07
Issue #261 fix

### LINKED ISSUE #261: Dependabot update fails with "pip in for packagename"
Example log: https://github.com/astropy/specreduce/actions/runs/14307324044

Don't know what this is about.

hpparvi:
This PR removes the `requirements.txt` file from the `notebook_sandbox/horne_extract` directory containing an early Horne extraction example notebook. The file contained a malformed requirement (`astropy=5.0` instead of `astropy==5.0`) that was likely the cause of the dependabot failures reported in #261.

Fix #261

### DIFF
--- a/notebook_sandbox/horne_extract/requirements.txt
+++ b/notebook_sandbox/horne_extract/requirements.txt
@@ -1,3 +1,0 @@
-astropy=5.0
-matplotlib==3.5.0
-numpy==1.21.4


pllim (approved):
FWIW LGTM. Thanks for digging! I was just generally confused because your dependabot was not even set up to check pip, so not sure where that check came from. I think the only way to find out is to merge and see?

PR #271, opened 2025-06-10
Drop python 3.10 support

hpparvi:
This PR removes support for Python 3.10, sets Python 3.11 as the minimum required version, and adds tests for Python 3.13. It also sets pytest to ignore two deprecation warnings from gwcs and specutils in devdeps tests (these are not related to specreduce), and enables manual workflow dispatch for "Python Tests" in GitHub. Finally, it updates the oldest supported dependencies for some packages to align with Scientific Python [SPEC 0](https://scientific-python.org/specs/spec-0000/).


.run/Template Python tests.run.xml
pllim, line 1: What is this for?

tox.ini
```diff
@@ -43,10 +42,10 @@ deps =
     devdeps: photutils>=0.0.dev0
     devdeps: git+https://github.com/spacetelescope/synphot_refactor.git#egg=synphot
 
-    oldestdeps: numpy==1.22.*
+    oldestdeps: numpy==1.24.*
```
pllim, line 45: While you are at it, might also want to sync this with minversion pins in pyproject.toml ?

pllim: Bumping Python and dep versions probably needs a change log as well. Thanks!

tox.ini
```diff
@@ -43,10 +42,10 @@ deps =
...
     oldestdeps: astropy==5.1.*
```
pllim, line 46: Are you sure you still want to support astropy 5.x ? That is pretty old.

Issue #276, opened 2025-06-26
Allow FitTrace to estimate the trace model using a subset of the whole dispersion-axis range (windowing in dispersion axis)


hpparvi:
To improve trace fitting by excluding low-SNR regions, FitTrace needs an option to crop the image along both the dispersion and cross-dispersion axes. The trace created from this cropped image should, however, still extend across the entire original image.

PR #277, opened 2025-06-28
Issue #276

hpparvi:
This PR addresses issue #276 by adding a ``disp_bounds`` argument to ``FitTrace``. The argument allows for adjusting the dispersion-axis window from which the trace peaks are estimated. The primary motivation behind this is to make life easier for end-users when tracing a spectrum that would otherwise require cropping.


tepickering (approved):
looks good! thanks for the fixes.

Issue #282, opened 2025-07-30
Include error propagation to spectrum extraction


hpparvi:
Error propagation needs to be included into 
 
- `BoxcarExtract.__call__`
- `HorneExtract.__call__` 
- `HorneExtract._fit_gaussian_spatial_profile`

PR #286, opened 2025-08-14
Add error propagation estimates for BoxcarExtract and HorneExtract

harry353:
This changes how uncertainties are handled in:

* `BoxcarExtract.__call__`
* `HorneExtract.__call__`
* `HorneExtract._fit_gaussian_spatial_profile`

as described in #282. Now they:

* Work correctly with both standard deviation and variance uncertainties.
* Pass the right uncertainty through to the extracted 1D spectrum.
* Keep units consistent.
* Handle NaNs and masked pixels better when fitting the Gaussian profile.

Also, I have added test scripts under the `tests` directory, as well as a .ipynb file for the Horne case with some examples.

### DIFF
--- a/notebook_sandbox/horne_extract/horne_uncertainty_testing.ipynb
+++ b/notebook_sandbox/horne_extract/horne_uncertainty_testing.ipynb
@@ -1,0 +1,326 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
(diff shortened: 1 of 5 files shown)


harry353:
Also currently working on issue #281 as well

hpparvi (changes requested):
Promising work @harry353! Here are a couple of initial comments. Can you ensure that the tests you've written run without errors and check also the code style (I'd recommend running your code through black).

PR #295, opened 2026-01-09
Boxcar extraction error propagation

### LINKED ISSUE #282: Include error propagation to spectrum extraction

hpparvi:
This PR adds uncertainty propagation to BoxcarExtract so that uncertainties from 2D input images are properly propagated to the extracted 1D spectrum. The approach supports `VarianceUncertainty`, `StdDevUncertainty`, and `InverseVariance` uncertainty types, and the PR includes a full set of tests for the functionality.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,10 @@
-1.7.1 (2025-11-2x)
+1.8.0 (2026-01-xx)
 ------------------
+
+New Features
+^^^^^^^^^^^^
+
+- Added uncertainty propagation to ``specreduce.extract.BoxcarExtract``.
 
 Other changes
 ^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


tepickering:
it looks like this and #296 have overlapping changes. at first glance it looks like #296 is based off the changes here which means either this should be merged first or just ignore this one and do everything in #296. it also looks like this and #296 mostly supercedes #286. we will want to reconcile that and see what we can still incorporate from there.

tepickering (approved):
my bad not looking close enough to see the changes were actually not overlapping. the way github displayed the diffs made them look so similar. which implies there may be some redundancy that could be refactored out at some point. at any rate, these changes are good to go so merge at will.

PR #296, opened 2026-01-11
Horne extraction error propagation

hpparvi:
This PR adds uncertainty propagation to HorneExtract so that uncertainties from 2D input images are properly propagated to the extracted 1D spectrum. Turns out this was very straightforward since we're already calculating the uncertainty for the Horne extraction itself (it's ``1/den`` without the normalisation). So, aside from handling uncertainty types and such, we actually needed to add one extra line of code... I also renamed ``img`` to ``flux`` so that the variable is not mixed with ``image``, and so that the variable naming is more understandable at the end of the method.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@
 New Features
 ^^^^^^^^^^^^
 
-- Added uncertainty propagation to ``specreduce.extract.BoxcarExtract``.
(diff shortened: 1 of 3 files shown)


tepickering (approved):
thanks for adding this! looks good except for one case that isn't covered by the new tests according to codecov. i would add that and then feel free to merge.

PR #301, opened 2026-02-16
V1.8

hpparvi:
This PR updates the changelog to cover the final changes between specreduce 1.7 and 1.8 and prepares specreduce for the v1.8 release. It also updates the PyPI publishing workflow to use Trusted Publishing, as recommended by @tepickering in #298.


tepickering (approved):
looks good! thanks for setting this up.

PR #303, opened 2026-03-15
2D Tilt Correction

hpparvi:
This PR introduces `specreduce.tilt_correction.TiltCorrection` and `specreduce.tilt_solution.TiltSolution` classes. The `TiltCorrection` class can be used to estimate the 2D tilt distortion as an astropy `Polynomial2D` model from (optionally multiple) arc frame(s), while the `TiltSolution` class can be used to access the tilt-corrected to detector mapping and its inverse, resample science frames, or export the tilt model as a `GWCS` object.

LLM usage: I've used Claude Code (Opus 4.6) to revise the language in the documentation, revise the docstrings, find and fix several bugs, improve the documentation of some parts of the code, and to implement the `TiltSolution._calculate_inverse` method (that I had implemented myself earlier, then removed since I considered it unnecessary, until I realized it may be quite necessary indeed in the future, but was too lazy to find the commit where I still had it). Additionally, I used Grammarly to fix the grammar in this PR description.


specreduce/tilt_solution.py
```diff
@@ -0,0 +1,369 @@
...
+            A compound model mapping detector coordinates to rectified dispersion coordinates.
+        """
+        image_shape = image_shape or self._image_shape
+        if image_shape is None:
+            raise TypeError(
+                "image_shape must be provided either at construction or as an argument."
+            )
+
+        ny, nx = image_shape
+        degree = degree or self._c2d[-1].degree + 1
```
cshanahan1, line 156: the docstring says this defaults to forward degree + 3

cshanahan1 (commented):
-

specreduce/tilt_solution.py
hpparvi: Good catch! I changed this after writing the docstring... Fixed now.

cshanahan1 (approved):

PR #310, opened 2026-05-26
Add missing pytest.mark.remote_data

olebole:
These were missing in 1.8.0; I discovered them when packaging the new version for Debian.


tepickering (approved):
good catch! thanks for fixing this oversight.

hpparvi:
Thanks @olebole, this was a good catch indeed! Merged.

PR #314, opened 2026-07-30
RTD annotation fix + tox action v2 -> v3 fix + linkcheck fix

hpparvi:
This small PR fixes the RDT build broken by the introduction of NumPy 2.5, the coverage tests broken by the transition from the tox v2 action to v3, and the weekly linkcheck action that was missing dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ test = [
     "tox",
 ]
 docs = [
+    "numpy<=2.5",
     "sphinx-astropy[confv3]>=1.11",
     "sphinx-copybutton",
     "sphinx-design",
--- a/.github/workflows/cron-tests.yml
+++ b/.github/workflows/cron-tests.yml
@@ -29,6 +29,10 @@ jobs:
       envs: |
         - name: Check URLs in docs
           linux: linkcheck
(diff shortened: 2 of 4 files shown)


cshanahan1 (approved):

pyproject.toml
```diff
@@ -24,6 +24,7 @@ test = [
     "tox",
 ]
 docs = [
+    "numpy<=2.5",
```
pllim, line 27: Why the max pin here?

hpparvi: The RTD build failed because ``nitpick_ignore`` in the Sphinx conf was tailored for Numpy < 2.5. Numpy 2.5 changed typing only slightly, but that was still enough to break the ignore rules. I've now tested that RTD builds the docs correctly with Numpy 2.5 (with these changes), and would prefer to pin the version so the docs build even if something similar happens in the future. We can update the pinned version as new Numpy versions are released, but only after first testing that the docs build fine.

pllim: I didn't know nitpick ignore is so... nitpicky. Maybe at least open a follow up issue and add that as a comment here so it does not get forgotten? Thanks!

PR #316, opened 2026-08-27
Spectral FITS WCS

hpparvi:
This PR adds:
- The `WavelengthSolution1D.wcs` method to calculate a spectral FITS WCS by fitting the grism dispersion function (Greisen et al. 2006) to the native polynomial solution. The WCS is returned as an `astropy.wcs.WCS` object that can be easily serialised to a FITS header.
- The `WavelengthSolution1D.attach_wcs` method to attach a (cached) FITS WCS to a `Spectrum` object, after which the spectrum can be saved/read using the  "wcs1d-fits" format.

AI/LLM disclaimer: I used Claude Opus 5 and Claude Fable quite extensively during development (including in agentic mode). That said, I am well familiar with the code and can explain what it does and why.

### DIFF
--- a/docs/wavelength_calibration/wavelength_calibration.rst
+++ b/docs/wavelength_calibration/wavelength_calibration.rst
@@ -214,9 +214,9 @@ Several tools help assess the quality of the wavelength solution:
 ********************************
 
 The fitted polynomial wavelength solution is stored as a
-:class:`~specreduce.wavesol1d .WavelengthSolution1D` instance that you can use to transform and
-resample spectra. The fitting methods return the solution, but it is also stored in
-`WavelengthCalibration1D.solution`.
+:class:`~specreduce.wavesol1d.WavelengthSolution1D` instance that you can use to transform and
+resample spectra. The fitting methods return the solution, but it is also stored in the
+``solution`` attribute of :class:`~specreduce.wavecal1d.WavelengthCalibration1D`.
 
 *   **Convert Coordinates**: Use :meth:`~specreduce.wavesol1d.WavelengthSolution1D.pix_to_wav` and
     :meth:`~specreduce.wavesol1d.WavelengthSolution1D.wav_to_pix` to convert between pixel and
--- a/docs/wavelength_calibration/wavelength_calibration.rst
+++ b/docs/wavelength_calibration/wavelength_calibration.rst
@@ -213,9 +213,10 @@ Several tools help assess the quality of the wavelength solution:
 5. Using the Wavelength Solution
 ********************************
 
-The fitted wavelength solution is stored as a :class:`~specreduce.wavesol1d.WavelengthSolution1D`
-instance that you can use to transform and resample spectra. The fitting methods return the
-solution, but it is also stored in ``WavelengthCalibration1D.solution``.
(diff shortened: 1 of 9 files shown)


tepickering:
i suspect the RTD build failure here is the same as with the other two recent PRs...

hpparvi:
The two PRs had slightly different issues in their docs, but both should be fixed now.

Also, I agree with dropping support for Python 3.11. I've now made Python 2.12 as the minium in these two PRs, and marked this to the changelog for specreduce 1.10 (the next release).

docs/wavelength_calibration/wavelength_calibration.rst
```diff
@@ -227,9 +228,23 @@ solution, but it is also stored in ``WavelengthCalibration1D.solution``.
...
+    software, or attached to a :class:`~specutils.Spectrum` object. A warning
+    is emitted if the fit deviates from the polynomial solution by more than ``max_residual``
+    pixels.
+
+    .. code-block:: python
+
+        header = ws.wcs().to_header()
+        print(header)
+
+*   **Export a GWCS Object**: Access the `~gwcs.wcs.WCS` object representing the polynomial
```
tepickering, line 244: might be useful to note here that GWCS maintains the full fidelity of the wavelength solution vs using an approximation to work around the constraints of the FITS standard.

specreduce/wavecal1d.py
```diff
@@ -145,7 +145,9 @@ def __init__(
 
         wave_air
             Boolean indicating whether the input wavelengths correspond to air rather than vacuum;
-            by default `False`, meaning vacuum wavelengths.
+            by default `False`, meaning vacuum wavelengths. The flag is also stored in the
+            wavelength solution, where it selects between the 'AWAV-GRA' and 'WAVE-GRI' axis
```
tepickering, line 149: someone looking at this who is not familiar with Paper III (like me until i went back and read it) would think `GRA` refers to grating and `GRI` refers to grism. even the paper itself makes it confusing. should make it clear here or in other documentation that the FITS standard uses `GRA` to refer to a disperser working in air wavelengths, and `GRI` to one working in vacuum.

specreduce/wavesol1d.py
```diff
@@ -38,12 +42,155 @@ def _diff_poly1d(m: models.Polynomial1D) -> models.Polynomial1D:
     return models.Polynomial1D(m.degree - 1, **coeffs)
 
 
+def _make_grism_wcs(
```
tepickering, line 48: i would personally say `disperser` instead of `grism`, but it does make sense to follow the terminology and conventions of the paper.

tepickering (approved):
looks good! actually using some of the spectroscopy functionality that FITS has long supported has been talked about for a while so i'm glad it's finally happening.

specreduce/wavecal1d.py
hpparvi: Very true, this was confusing to me as well. I've now expanded the explanation in the docstring.

docs/wavelength_calibration/wavelength_calibration.rst
hpparvi: True, fixed!

PR #317, opened 2026-08-27
1D wavelength solution ASDF export/import

hpparvi:
This PR adds the `WavelengthSolution1D.to_asdf` and `WavelengthSolution1D.from_asdf` methods to serialize a 1D wavelength solution into an ASDF file. 

AI/LLM disclaimer: similarly to #316, I used Claude Fable to speed up the development.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ New Features
 - Added ``WavelengthSolution1D.to_asdf()`` and ``WavelengthSolution1D.from_asdf()``
   for serializing a wavelength solution losslessly into an ASDF file. Only the
   coordinate transformation is stored, as a GWCS object whose bounding box carries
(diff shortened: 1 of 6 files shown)


tepickering:
i will give this a more thorough review later this week. the python 3.11 and RTD failures should be looked into. i would argue that we can drop 3.11 at this point.

hpparvi:
Thanks, @tepickering! I've fixed the RDT issues and dropped support for Python 3.11.

tepickering (approved):
GWCS and ASDF are the way forward. need to get the rest of the community to come along...

cshanahan1 (approved):

tepickering:
looks like we missed pruning the python 3.11 test from the CI so it's failing. good timing on dropping it, though, because astropy will drop it as well as of v8.1.

PR #319, opened 2026-09-05
WavelengthSolution1D.resample bug fixes and improvements

hpparvi:
This PR fixes several bugs in the `WavelengthSolution1D.resample` method and improves the method's overall functionality.

Bug fixes                                                                                                                                                                                                                                                                                                                                     
- The flux was multiplied by the pixel wavelength width before dividing by the bin width, so the output was labeled as a flux density but was off by the local dispersion and did not conserve the integrated flux. The density and its uncertainty are now computed solely from the fractional pixel overlaps, with the variance using the squared weights. (I somehow remembered I had fixed this a long time ago, so I'm not sure what happened here...)
- Spectra whose spectral axis does not start at pixel zero were indexed from the wrong pixels. 
- Solutions where the wavelength decreases with pixel number silently produced zero flux. The output is now always ascending in wavelength.

Changes 
- The input mask is propagated (a bin is flagged if any contributing pixel is masked) and the metadata is copied. 
- A spectrum without uncertainty returns no uncertainty instead of a fabricated zero-valued one.
- The output spectral axis carries the bin edges actually used, so `spectral_axis.bin_edges` is exact for non-uniform grids and integrating the flux over those edges recovers the input counts exactly.

AI and LLM use disclaimer: I'm systematically going through the specreduce codebase with Claude Code (Fable 5.1) to identify bugs, issues, and any room for (non-breaking) improvements in the existing functionality before the specreduce 2.10 release. The first, rather serious, bug I was aware of beforehand, but I remembered I had it already fixed several months ago. I also used Claude Code to implement most of the changes. I understand the code and the changes and could have implemented them myself, but it would have taken significantly longer than the 2-3 hours of continuous work it took to create this PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,9 +33,13 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
-- Fixed the variance propagation in ``WavelengthSolution1D.resample()``, which
-  weighted each pixel's variance by its wavelength width instead of the square of the
-  width, underestimating the uncertainty of the resampled flux. [#XXX]
+- Fixed ``WavelengthSolution1D.resample()``, which multiplied the per-pixel flux by
+  the wavelength width of each pixel before dividing by the bin width. The output was
+  labelled as a flux density per wavelength unit but was numerically off by the local
+  dispersion, so a flat spectrum in counts per pixel stayed flat instead of falling where
+  the dispersion grows, and the integrated flux was not conserved. The flux density and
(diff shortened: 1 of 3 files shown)


tepickering:
looks like python 3.12 is now required, but the CI not updated to remove 3.11. the windows failure is the old tcl bug. i _think_ that can be fixed by forcing matplotlib to use the `Agg` head-less backend in the test fixtures.

PR #321, opened 2026-09-07
find_arc_lines uncertainty handling fix

hpparvi:
This PR fixes two issues in how the `find_arc_lines` handles uncertainties:

1. the `find_arc_lines` function now allows uncertainties to be any of Astropy's uncertainty types (it allowed previously only `StdDevUncertainty`)

2. If the spectrum doesn't include uncertainties, the uncertainties are estimated using a robust approach based on the DER_SNR algorithm by [Stoerh et al. (2008)](https://adsabs.harvard.edu/pdf/2008ASPC..394..505S). (with a small difference that we're using Astropy's sigma-clipped median instead of plain median).

I've also added the noise estimation function as `specreduce.utils.utils.measure_noise`.

AI/LLM Disclaimer: The same as in #319. I've used Claude Code with Fable 5.1 to co-develop the PR, but I understand what the code does and how it works, and can explain it when needed.


tepickering:
python 3.11 still in CI when 3.12 is minimum required.

PR #322, opened 2026-09-07
Fix and improve HorneExtract

hpparvi:
This PR fixes a number of accuracy issues in `HorneExtract` and makes it follow the trace at sub-pixel precision.

### Bug fixes

  - **Non-flat traces**: the flux was rolled to align the trace with the central row, but the variance and mask arrays were not, so the extraction weights came from the wrong pixels. This biased the flux and its uncertainty when variances were non-uniform or when pixels were masked.

  - **Low flux bias with data-derived variances**: weighting by the Poisson variance of the observed counts underweights upward fluctuations and biases the flux low by roughly the inverse of the counts per pixel. Following Horne, the variances are now re-estimated from the extraction model before the final extraction. The new `model_variance` argument (default `True`) controls this.

  - **Extraction kernel**: the kernel is now the fitted Gaussian alone. The background model no longer leaks into the kernel and its normalisation, which could drive the flux towards zero far from the source.

  - **`bkgrd_prof=None`** now fits the profile without a background model instead of silently substituting the default polynomial.

### Improvements

  - **Sub-pixel trace following**: the kernel is evaluated directly on the image grid at each pixel's offset from the trace instead of rolling the image by whole pixels. The profile is fit in trace-relative coordinates without resampling (Horne 1986, Sect. II.A).
                                                                                                                                                                                                                                                                                               
  - **`window` argument** restricts the profile fit and extraction to pixels near the trace, keeping other sources on the slit out of the fit.
  
  - **Vectorised the kernel evaluation**, which makes the extraction about three times faster.

AI/LLM Disclaimer: The same as in #319. I've used Claude Code with Fable 5.1 to co-develop the PR, but I understand what the code does and how it works, and can explain it when needed.


tepickering:
looks like CI is only passing under python 3.14.

hpparvi:
Ohps, true... I'll take a look into this tomorrow morning.

tepickering:
also, see my other comments about 3.11 still in CI when 3.12 is now the minimum. should fix that in a separate PR and then rebase these once that's merged.

PR #323, opened 2026-09-17
Remove a missed Python 3.11 tox test

hpparvi:
This PR removes a Python 3.11 tox test that I missed in an earlier PR. After this, we have cleanly dropped support for Python 3.11.


tepickering (approved):
thank you! python 3.11 served us well, but it's time to move on. if we want to follow SPEC0, https://scientific-python.org/specs/spec-0000/, then python 3.12's time will be up soon as well.

